Issue #43, opened 2017-11-03
runtime error - MacOS


tbatch:
$ aws-nuke-v1.2.1-darwin-amd64 -c aws-nuke-config.yml --profile aws-enterprise-uat

us-east-1 - RDSDBSubnetGroup - 'apiture-stacks-component-rds-databasesubnetgroup-17hxceov4kkcc' - would remove
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x185dbf6]

goroutine 53 [running]:

svenwltr:
@tbatch Could you also post the stack trace, which comes after that error?

tbatch:
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x185dbf6]

goroutine 29 [running]:
github.com/rebuy-de/aws-nuke/resources.(*S3Nuke).DescribeBuckets(0xc420074240, 0x244fda0, 0x0, 0x0, 0x0, 0x0)
	/go/src/github.com/rebuy-de/aws-nuke/resources/s3-buckets.go:28 +0x166
github.com/rebuy-de/aws-nuke/resources.(*S3Nuke).ListBuckets(0xc420074240, 0x244fda0, 0x0, 0x0, 0x0, 0x0)
	/go/src/github.com/rebuy-de/aws-nuke/resources/s3-buckets.go:38 +0x40
github.com/rebuy-de/aws-nuke/resources.(*S3Nuke).ListBuckets-fm(0x244fda0, 0x0, 0x0, 0x0, 0x0)
	/go/src/github.com/rebuy-de/aws-nuke/resources/listers.go:101 +0x2a
github.com/rebuy-de/aws-nuke/cmd.Scan.func1(0xc420138180, 0xc4204d0320, 0xc42019eae0)
	/go/src/github.com/rebuy-de/aws-nuke/cmd/scan.go:22 +0x77
created by github.com/rebuy-de/aws-nuke/cmd.Scan
	/go/src/github.com/rebuy-de/aws-nuke/cmd/scan.go:39 +0x9e

tbatch:
Sorry about that, the stack trace is above.

tbatch:
Maybe something wrong with my region?  Below is my config file.

regions:
- us-east-1

account-blacklist:
- "0000000000" # production - real account number here.

accounts:
[...]

PR #44, opened 2017-11-03
fix another null pointer dereference in S3

svenwltr:
> https://github.com/rebuy-de/aws-nuke/issues/43

I am not sure in which cases one of the two pointers might be null.

### DIFF
--- a/resources/util.go
+++ b/resources/util.go
@@ -8,3 +8,15 @@ import (
 func GetCategory(r Resource) string {
 	return strings.Split(fmt.Sprintf("%T", r), ".")[1] // hackey
 }
+
+func EqualStringPtr(v1, v2 *string) bool {
+	if v1 == nil && v2 == nil {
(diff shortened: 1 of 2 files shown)


resources/util.go
```diff
@@ -8,3 +8,15 @@ import (
 func GetCategory(r Resource) string {
 	return strings.Split(fmt.Sprintf("%T", r), ".")[1] // hackey
 }
+
+func EqualStringPtr(v1, v2 *string) bool {
+	if v1 == nil && v2 == nil {
+		return true
+	}
+
+	if v1 != nil || v2 != nil {
```
bethge: Shouldn't this be: `v1 == nil || v2 == nil` ?

Iiuc it would now always return `false` if at least one of the two pointers is not equal to `nil`.

svenwltr: true. fixed

PR #48, opened 2017-11-20
CLOUD-1310 adds rds cluster-related resources

octopusx:
This commit enables aws-nuke to recognise DBClusterParameterGroup and
DBCluster type resources and remove them.
The changes are necessary to nuke Aurora clusters.

@rebuy-de/prp-aws-nuke please review


resources/rds-dbclusterparametergroups.go
```diff
@@ -0,0 +1,56 @@
...
+	return nil
+}
+
+func (i *RDSDBClusterParameterGroup) Remove() error {
+	params := &rds.DeleteDBClusterParameterGroupInput{
+		DBClusterParameterGroupName: i.name,
+	}
+
+	_, err := i.svc.DeleteDBClusterParameterGroup(params)
+	if err != nil {
```
bethge: Couldn't you just skip this `if` and directly: `return err`

stephanlindauer: IMHO the logic is more understandable and separated like this...
first: error handling
then: making the function return what is should

svenwltr: IMO it looks like unnecessary code :>

octopusx: You could, but then you are implying that whoever looks at the code afterwards understands that `err` defaults to `nil`. The if statement makes it more explicit and easier to read what's going on. Also, this is how that error is handled across all the classes in aws-nuke where it's applicable, so it is following a common pattern in the program.

octopusx: Decision?

svenwltr: Your call

octopusx: I will leave it in for consistency's sake. If this is a big enough problem then it deserves its own pull request and that if statement should be removed everywhere across the whole application.

bethge: Very much agree. Greatly in favor of consistency.

PR #55, opened 2017-12-02
Golang capitalization

bethge:
@rebuy-de/prp-aws-nuke - PTAL.

Changing to consistent golang-based capitalization for all resource types. I hope this is everything.

Once reviewed, I would squash the commits before merging via merge commit.


svenwltr:
I don't think you need to squash that, since the commits make sense on their own.

PR #56, opened 2017-12-02
Resolve deprecated types

bethge:
@rebuy-de/prp-aws-nuke - PTAL.

This PR is based on: https://github.com/rebuy-de/aws-nuke/pull/55

Not perfectly happy with it, would much appreciate feedback. In particular:
- The list of deprecations in the function is clumsy
- Logging of the warning is not great. Better logging would require changes to the existing logging facilities


cmd/config.go
```diff
@@ -46,3 +51,46 @@ func (c *NukeConfig) InBlacklist(searchID string) bool {
...
+		"IamUser":                  "IAMUser",
+	}
+
+	for _, a := range c.Accounts {
+		for resourceType, resources := range a.Filters {
+			replacement, ok := deprecations[resourceType]
+			if !ok {
+				continue
+			}
+			fmt.Printf("deprecated resource type '%s' - replace with '%s'\n", resourceType, replacement)
```
svenwltr: You could either do `LogError(fmt.Errorf("depe...` or actually change `LogError` so we have something like `LogError` and `LogErrorString`.

svenwltr:
Maybe we also should consider ignoring the case of the types in the config. What do you think about this @bethge?

bethge:
I am mulling it over. Certainly fewer resource types would need renaming and being bothered to modify configurations just for character casing reasons is rather nitpicky.

On the other hand the cases of `*Attachement => *Attachment` and `IamUserAccessKeys => IamUserAccessKey` would remain.

Looking at k8s, they seem very strict about casing, but that is likely due to cross-API compatibility. Given the chance to start from zero, I would go with golang camelcasing.

Tbh though, as long as we don't break anyone's config I am ok with it either way.

svenwltr:
So you have a light preference towards strict casing?

> --
Sven Walter
Software Developer

[image:
https://docs.google.com/uc?export=download&id=0B66Nk4FC03uxR0hkR2tTYVVVZ3M&revid=0B66Nk4FC03uxajBwR3M5d2UvSU1Oa1BybUcxNWIyMUcvQlV3PQ]

reBuy reCommerce GmbH | Potsdamer Str. 188 | 10783 Berlin

*mail: s.walter@rebuy.com <s.walter@rebuy.com> *| *web*: www.reBuy.de
<http://www.rebuy.de/>



*Geschäftsführer*: Torsten Schero

*Sitz und Registergericht*: Amtsgericht Charlottenburg, HRB 109344 B,
Ust-IdNr. DE237458635

bethge:
Yeah - unless someone complains about unnecessary casing changes, I would go with strict casing.

cmd/config.go
svenwltr: @bethge I thought about this. I would love a yellow-ish warning message :-) (ie `LogWarning`)

PR #57, opened 2017-12-02
Fix capitalization of "IamUser"

clebio:
This seems to matter. With `IAMUser`, my user doesn't get filtered out. With `IamUser`, it does.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ account-blacklist:
 accounts:
   "000000000000": # aws-nuke-example
     filters:
-      IAMUser:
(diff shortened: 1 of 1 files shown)


svenwltr:
Hello @clebio. Thanks for your PR. The readme has a wrong capitalization, because we are [currently aligning](https://github.com/rebuy-de/aws-nuke/pull/52) the different capitalization in the resource names and we didn't release the changes yet. I am very sorry for that confusion.

We will do a release in the next couple of days, which allows both versions and prints a proper warning message.

PR #60, opened 2017-12-04
CLOUD-1332 adds dynamodb items nuking

octopusx:
- DynamoDB Tables are now removed
- DynamoDB Table Items are now removed

@rebuy-de/prp-aws-nuke please review

### DIFF
--- a/resources/dynamodb-items.go
+++ b/resources/dynamodb-items.go
@@ -1,0 +1,80 @@
+package resources
+
+import (
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+)
+
+type DynamoDBTableItem struct {
+	svc   *dynamodb.DynamoDB
+	id    map[string]*dynamodb.AttributeValue
+	table Resource
+}
+
+func (n *DynamoDBNuke) ListItems() ([]Resource, error) {
+	tables, tablesErr := n.ListTables()
+	if tablesErr != nil {
+		return nil, tablesErr
+	}
+
+	resources := make([]Resource, 0)
+	for _, dynamoTable := range tables {
+		describeParams := &dynamodb.DescribeTableInput{
+			TableName: aws.String(dynamoTable.String()),
+		}
+
+		descResp, descErr := n.Service.DescribeTable(describeParams)
+		if descErr != nil {
+			return nil, descErr
+		}
+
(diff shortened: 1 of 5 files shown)


resources/dynamodb-items.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (i *DynamoDBTableItem) String() string {
+	table := i.table.String()
+	keyField := ""
```
bethge: afaik you could use: `var keyField string`, but it is purely aesthetics.

bethge:
I wonder if we should add some kind of warning that blacklisting a DynamoDB table would still delete all its items. Though maybe the pre-nuke output is enough.

resources/dynamodb-items.go
```diff
@@ -0,0 +1,81 @@
...
+
+	for _, value := range i.id {
+		value := strings.Replace(value.String(), "\n", " ", -1)
```
svenwltr: You could just use `strings.TrimSpace()` here.

resources/dynamodb-items.go
```diff
@@ -0,0 +1,81 @@
...
+		keyFieldChars := []rune(value)
+		keyField = string(keyFieldChars[8:(len(keyFieldChars) - 3)])
```
svenwltr: Slicing around with strings really looks hacky. Also it's hard to understand, since I have no idea how the contents look like.

So why do you need to do that?

Does this help? https://docs.aws.amazon.com/sdk-for-go/api/service/dynamodb/dynamodbattribute/#ConvertFrom

octopusx:
Applied minor updates to address the pull request comments
@rebuy-de/prp-aws-nuke please review

resources/dynamodb-tables.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+)
+
+type DynamoDBTable struct {
+	svc *dynamodb.DynamoDB
+	id  string
+}
+
+func (n *DynamoDBNuke) ListTables() ([]Resource, error) {
+	params := &dynamodb.ListTablesInput{}
```
stephanlindauer: this could be 
```
resp, err := n.Service.ListTables(&dynamodb.ListTablesInput{})
```
because `params` is not used in other places.

PR #61, opened 2017-12-05
Rename rds resouce

bethge:
@rebuy-de/prp-aws-nuke - PTAL

* RDSCluster -> RDSDBCluster


bethge:
It never made it into production, but are right, better be safe than sorry.

Issue #64, opened 2017-12-12
Delete all versions in S3 bucket


sabrehagen:
Deleting versioned `S3` buckets fails if all versions are not deleted prior to bucket deletion:

```
us-east-2 - S3Bucket - 's3://application-state' - BucketNotEmpty: The bucket you tried to delete is not empty. You must delete all versions in the bucket.
	status code: 409, request id: 166516D767C46F79, host id: xezvJzBCa84wcRsXRndei9W8Vz9iqDgZZ/Yad5LZfMQL2i9p4LRNfu8wLcR9AMWWCsMJf1A+Sdk=
```

svenwltr:
Duplicate of #62.

Issue #62, opened 2017-12-05
Delete all versions of s3_object


stephanlindauer:
When `aws-nuke` wants to delete a bucket with versioned `s3_object`s inside, it yields this error message:
```
BucketNotEmpty: The bucket you tried to delete is not empty. You must delete all versions in the bucket.
	status code: 409, request id: XXXXXX, host id: XXXXXX
```
`aws-nuke` should be able to remove those versioned `s3_object`s.

PR #66, opened 2017-12-14
Refactor resource listers

svenwltr:
Some heavy refactoring in preparation of #54. Also I want to get rid of the weird reflections.

Since this is a heavy change, here some notes:

This changes the interface for registering resources. Previously it was necessary to update the `GetListers` function in `resources/listers.go`. Now every resource registers itself inside of an `init` function which is defined in `resources/interface.go` (eg `register("IAMUser", ListIAMUsers)`). Since it has to provide a name for registering it isn't required anymore to get the resource type via reflections.

Also we are getting rid of the `FooNuke` types in `resources/types.go`, which just carried the session and don't really give a benefit. Now, the `ResourceLister` interface changed so it takes a session as a command. Therefore the intermediate types aren't required anymore.

PR #67, opened 2017-12-19
Adds support for s3 object versions

simonswine:
* Use ListObjectVersions instead
* Ensure that you can retrieve objects in buckets bigger than 1000
Objects/Versions

This should address  issues #62 #64

### DIFF
--- a/resources/s3-objects.go
+++ b/resources/s3-objects.go
@@ -7,9 +7,10 @@ import (
 )
 
 type S3Object struct {
-	svc    *s3.S3
-	bucket string
-	key    string
+	svc       *s3.S3
+	bucket    string
+	key       string
+	versionID *string
 }
 
 func (n *S3Nuke) ListObjects() ([]Resource, error) {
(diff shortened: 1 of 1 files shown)


resources/s3-object-versions.go
```diff
@@ -0,0 +1,33 @@
+package resources
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/service/s3"
+)
+
+type S3ObjectVersion struct {
```
svenwltr: I wouldn't create a new type for that. Currently the `S3Nuke.ListObjects` produces two types which are `S3ObjectVersion` and `S3Object`.

The problem there is that the filtering and refreshing currently relies on weird reflections of the type. This would work for now, but I want to get rid of this reflection code and make things more explicit (#66).

This means that one type should return exactly one type. Otherwise filtering and refreshing would get complicated, because *aws-nuke* would need to scan all resources on every iteration. This would slow down the execution and cause an unnecessary usage of the AWS API.

So my suggestion would be to merge these two types and make the `versionID` optional (by converting it to a reference). This is basically what the AWS SDK already does.

resources/s3-objects.go
```diff
@@ -21,21 +21,57 @@ func (n *S3Nuke) ListObjects() ([]Resource, error) {
...
+					resources = append(resources, &S3Object{
+						svc:    n.Service,
+						bucket: name,
+						key:    *out.Key,
+					})
+				}
+			}
+
+			for _, out := range resp.DeleteMarkers {
+				if out.VersionId != nil && *out.VersionId != "null" {
```
svenwltr: According to the [Go conventions](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) it should be named `VersionID`.

stephanlindauer:
hi @simonswine, 
whats the current state of this PR from your side? 
i would volunteer to get the refinements in, in case you are too busy.

resources/s3-objects.go
simonswine: Just followed what AWS is doing: https://github.com/aws/aws-sdk-go/blob/master/service/s3/api.go#L7325

resources/s3-object-versions.go
simonswine: I am not too sure, to do this right can be quite tricky and I could image a couple of edge cases. That's why I decided to use a separate type for it.

Happy to give the integrated type a try, but won't guarantee for anything  :laughing:

simonswine:
@stephanlindauer thanks for the reminder, will have a look over the next couple of days, feel free to remind me :laughing:

simonswine:
OK just pushed the changes, only tested it with versioned buckets for now....

resources/s3-objects.go
```diff
@@ -21,21 +22,41 @@ func (n *S3Nuke) ListObjects() ([]Resource, error) {
...
+			resp, err := n.Service.ListObjectVersions(params)
+			if err != nil {
+				return nil, err
+			}
+
+			for _, out := range resp.Versions {
+				resources = append(resources, &S3Object{
+					svc:       n.Service,
+					bucket:    name,
+					key:       *out.Key,
```
bethge: Don't you need to add the `versionID` to the [`DeleteObjectInput`][1]?

[1]: https://github.com/rebuy-de/aws-nuke/pull/67/files#diff-14f2e0b0cebab6efa8bc9b13cbeca4bbR67

svenwltr:
I rebased it to master and changed it according to the comments from @bethge: #76

PR #70, opened 2018-01-05
Refactor AWS account validation

svenwltr:
* move AWS account queries into `awsutil` package to slim down `cmd` package
* move account validation from `Nuke` to the actual config struct
* add tests for account validation

This is a preparation for #66.

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -53,38 +53,37 @@ func (c *NukeConfig) InBlacklist(searchID string) bool {
 	return false
 }
 
-func (c *NukeConfig) ValidateAccount(accountID string, aliases []string) (*NukeConfigAccount, error) {
+func (c *NukeConfig) ValidateAccount(accountID string, aliases []string) error {
 	if !c.HasBlacklist() {
-		return nil, fmt.Errorf("The config file contains an empty blacklist. " +
+		return fmt.Errorf("The config file contains an empty blacklist. " +
 			"For safety reasons you need to specify at least one account ID. " +
 			"This should be your production account.")
 	}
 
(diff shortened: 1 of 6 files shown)


cmd/config_test.go
```diff
@@ -142,3 +143,41 @@ func TestResolveDeprecations(t *testing.T) {
...
 	}
 }
+
+func TestConfigValidation(t *testing.T) {
+	cwd, err := os.Getwd()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	config, err := LoadConfig(path.Join(cwd, "..", "config", "example.yaml"))
```
bethge: Somewhat conflicted about using an example file as a test input. It does make sure that the example is valid to some degree. On the other hand it is really unexpected to have an example be part of a test case.

I do not have much experience with generated documentation, but I feel like this would be another potential use case for it (besides the documentation of supported AWS resources).

cmd/config.go
```diff
@@ -51,6 +53,40 @@ func (c *NukeConfig) InBlacklist(searchID string) bool {
 	return false
 }
 
+func (c *NukeConfig) ValidateAccount(accountID string, aliases []string) (*NukeConfigAccount, error) {
```
bethge: Imho, I liked the old signature of `ValidateAccount`, where it only validates and returns error(s).

svenwltr: That's very true. Fixed.

cmd/config_test.go
svenwltr: I had the same doubts. I added another test config for the tests.

PR #71, opened 2018-01-09
Add IAMUserPolicy resource

stephanlindauer:
Please review @rebuy-de/prp-aws-nuke

### DIFF
--- a/resources/iam-user-policy.go
+++ b/resources/iam-user-policy.go
@@ -1,0 +1,64 @@
+package resources
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/iam"
+)
+
+type IAMUserPolicy struct {
+	svc        *iam.IAM
+	userName   string
+	policyName string
+}
+
+func init() {
+	register("IAMUserPolicy", ListIAMUserPolicies)
+}
+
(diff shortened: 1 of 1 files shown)


resources/iam-user-policy.go
```diff
@@ -0,0 +1,64 @@
...
+				policyName: *policyName,
+				userName:   *user.UserName,
+			})
+		}
+	}
+
+	return resources, nil
+}
+
+func (e *IAMUserPolicy) Remove() error {
```
bethge: Why `e` though?

stephanlindauer: its named like this in other places as well. for example: 
https://github.com/rebuy-de/aws-nuke/blob/2d9b2f8fe3937ae7972a9ccc41159470525778f7/resources/iam-users.go#L36
https://github.com/rebuy-de/aws-nuke/blob/2d9b2f8fe3937ae7972a9ccc41159470525778f7/resources/iam-user-policy-attachments.go#L52
i'm also not a big fan of those ambiguous one letter variable names but i felt like following the style of the other resources out-weights that urge in that case.
if we adapt to having proper variable naming here, then i'd vote for doing it in all the places at once in a separate pull request.

svenwltr: FYI https://github.com/golang/go/wiki/CodeReviewComments#receiver-names

PR #74, opened 2018-01-26
 Add resource filters to config

### LINKED ISSUE #54: Add resource filters to config
would be nice to have a filter for the targets [include/exclude] to get selective on which targest

```
AutoScalingGroup
CloudFormationStack
CloudTrailTrail
CloudWatchEventsRule
CloudWatchEventsTarget
EC2Address
EC2DhcpOption
EC2Instance
EC2InternetGateway
EC2InternetGatewayAttachement
EC2KeyPair
EC2NetworkACL
EC2RouteTable
EC2SecurityGroup
EC2Subnet
EC2Volume
EC2Vpc
EC2VpcEndpoint
ELBv2TargetGroup
KMSAlias
KMSKey
LambdaFunction
LaunchConfiguration
RDSDBSubnetGroup
SNSSubscription
SNSTopic
```

I was able to use a quick for loop and invoke via script as `aws-nuke -c config.yml --profile myuser --target $i --force --no-dry-run &` but would be nice to be able in the config.yml filter by service to nuke or not

svenwltr:
> Closes #54 

Unlike proposed in #54, I let the flag named `--target`. I think my confusion about union and intersection came from the flag name. By naming the flag target it should be more clear how it behaves.

Also I updated the command line flag descriptions.


README.md
```diff
@@ -196,6 +196,53 @@ for debugging, if the error is not about dependencies.
...
+      - S3Bucket
+```
+
+If targets are specified in multiple places (eg CLI and account specific), then
+a resource type must be specified in all places. In other words each
+configuration limits the previous ones.
+
+If a exclude is defined in any place, then if will be ignored in any case.
+
+**Hint:** You can see all available resources types with this command:
```
stephanlindauer: **resource** types

README.md
```diff
@@ -196,6 +196,53 @@ for debugging, if the error is not about dependencies.
...
+    resource-types:
+      exclude:
+      - S3Bucket
+```
+
+If targets are specified in multiple places (eg CLI and account specific), then
+a resource type must be specified in all places. In other words each
+configuration limits the previous ones.
+
+If a exclude is defined in any place, then if will be ignored in any case.
```
stephanlindauer: **an** exclude
**it** will be ignored

README.md
svenwltr: fixed

README.md
svenwltr: fixed

svenwltr:
bump

README.md
```diff
@@ -196,6 +196,53 @@ for debugging, if the error is not about dependencies.
 *aws-nuke* retries deleting all resources until all specified ones are deleted
 or until there are only resources with errors left.
 
+### Specifying Resource Types to Delete
+
+*aws-nuke* deletes a lot of resources and there might be added more at any
+release. Eventually, every resources should get deleted. You might to restrict
+which resources to delete. There are multiple ways to configure this.
+
+One way are filter, which already got mentioned. This requires to know the
```
bjoernhaeuser: `filters`

svenwltr: fixed

config/example.yaml
```diff
@@ -4,8 +4,18 @@ regions:
 account-blacklist:
 - 1234567890
 
+resource-types:
+  target:
```
bethge: Consider having two examples, one with a blacklist and another one with a whitelist.

cmd/config.go
```diff
@@ -5,17 +5,26 @@ import (
...
 	"strings"
 
+	"github.com/rebuy-de/aws-nuke/pkg/types"
+
 	"gopkg.in/yaml.v2"
 )
 
+type ResourceConfig struct {
+	Targets  types.Set `yaml:"targets"`
+	Excludes types.Set `yaml:"excludes"`
```
bethge: Shouldn't this be `yaml:"target"` and `yaml:"exclude"` ?

README.md
```diff
@@ -196,6 +196,53 @@ for debugging, if the error is not about dependencies.
...
+    resource-types:
+      exclude:
+      - S3Bucket
+```
+
+If targets are specified in multiple places (eg CLI and account specific), then
+a resource type must be specified in all places. In other words each
+configuration limits the previous ones.
+
+If an exclude is defined in any place, then it will be ignored in any case.
```
bethge: Consider being more explicit, along the lines of:
> If an exclude is used, then all its resource types will not be deleted.

cmd/config.go
svenwltr: IMO not, since you specify a list.

bethge: Makes sense. Doesn't this need to be `targets`: https://github.com/rebuy-de/aws-nuke/pull/74/files#diff-04c6e90faac2675aa89e2176d2eec7d8R223 ?

pkg/types/list.go
```diff
@@ -0,0 +1,39 @@
+package types
+
+type Set []string
```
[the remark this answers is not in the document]
svenwltr: Using a `map` also has these two downsides:

* We need a constructor, which currently isn't required (which I like).
* Since our lovely golang map gets shuffled all the time, we get non-deterministic results (ie resources get nuked in a different order every time).

README.md
svenwltr: fxied

cmd/config.go
svenwltr: fexid

config/example.yaml
svenwltr: fxeid

pkg/types/list.go
```diff
@@ -0,0 +1,39 @@
+package types
```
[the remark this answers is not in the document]
svenwltr: fiexd

pkg/types/list.go
```diff
@@ -0,0 +1,39 @@
...
+		if !m[t] {
+			result = append(result, t)
+		}
+	}
+
+	return result
+}
+
+func (s Set) Union(o Set) Set {
+	return Set(append(s, o...))
```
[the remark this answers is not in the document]
svenwltr: feixd

svenwltr:
bump

README.md
```diff
@@ -196,6 +196,67 @@ for debugging, if the error is not about dependencies.
 *aws-nuke* retries deleting all resources until all specified ones are deleted
 or until there are only resources with errors left.
 
+### Specifying Resource Types to Delete
+
+*aws-nuke* deletes a lot of resources and there might be added more at any
+release. Eventually, every resources should get deleted. You might to restrict
```
octopusx: "you might want to restrict" ?

pkg/types/list.go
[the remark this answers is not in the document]
svenwltr: With the current implementation it would remove both.

I am not eager to implement something just to match a name. We can rename it, if you suggest a better name.

README.md
svenwltr: fixed

octopusx: :+1:

pkg/types/list.go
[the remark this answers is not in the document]
bethge: Oh yeah, you are totally right, it would remove both. Sorry about that.

Hmm..., naming things is hard ... how about `LazySet`. What do you think? It would imply it is not quite a super strict set but it certainly is a set. Feel free to stick with `Set` if you prefer.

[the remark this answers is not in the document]
svenwltr: "lazy" always makes me think of lazy loading. What about `Collection` to remove any implication from the name?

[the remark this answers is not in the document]
bethge: Good idea! I like it.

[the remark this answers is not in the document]
svenwltr: renamed

PR #87, opened 2018-03-02
Feature/mq Brokers

tomvachon:
Add MQ Service


svenwltr:
You are using two GitHub accounts here. Is that on purpose?

tomvachon:
Forgot to set user. I set it now

PR #96, opened 2018-03-02
Feature/datapipeline

tomvachon:
Add DatePipeline Pipeline cleanup


tomvachon:
Can't tell why this failed, seems like an upstream go get hiccup

svenwltr:
A rebuild fixed it.

PR #100, opened 2018-03-05
Add Cloudwatch Dashboards and Alarms

tomvachon:
I know that the slice casting in the remove isnt ideal.  If you have a better way (and frankly the right way) to do it, let me know.

### DIFF
--- a/resources/cloudwatch-alarms.go
+++ b/resources/cloudwatch-alarms.go
@@ -1,0 +1,60 @@
+package resources
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/cloudwatch"
+)
+
+type CloudWatchAlarm struct {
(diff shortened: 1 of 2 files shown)


resources/cloudwatch-alarms.go
```diff
@@ -0,0 +1,60 @@
...
+		params.NextToken = output.NextToken
+	}
+
+	return resources, nil
+}
+
+func (f *CloudWatchAlarm) Remove() error {
+
+	_, err := f.svc.DeleteAlarms(&cloudwatch.DeleteAlarmsInput{
+		AlarmNames: []*string{f.alarmName},
```
bethge: I think this is good. The aws utils seem to be only for easy casting.

svenwltr: Actually, this is not a cast. It is just putting a single item into a slice which is perfectly fine.

tomvachon: OK, thanks! Glad to know I'm not too bad a grade school Go!

PR #109, opened 2018-03-06
Add Glue Service

tomvachon:


resources/glue-classifiers.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	for {
+		output, err := svc.GetClassifiers(params)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, classifier := range output.Classifiers {
+			if classifier.GrokClassifier != nil {
```
svenwltr: You still could use a switch here. Something like this:

```golang
switch {
case classifier.GrokClassifier != nil:
    resources = append(resources, &GlueClassifier{
        svc:  svc,
        name: classifier.GrokClassifier.Name,
    })
case classifier.JsonClassifier != nil:
    resources = append(resources, &GlueClassifier{
        svc:  svc,
        name: classifier.JsonClassifier.Name,
    })
case classifier.XMLClassifier != nil:
    resources = append(resources, &GlueClassifier{
        svc:  svc,
        name: classifier.XMLClassifier.Name,
    })
}
```

I like the switch a bit more here, because as a reader it is more obvious, that only one  block should be executed. Also note that Golang doesn't [fall through by default](https://github.com/golang/go/wiki/Switch#fall-through).

tomvachon: OK, ill make a change

PR #118, opened 2018-03-07
Add KinesisAnalytics Application

tomvachon:


resources/kinesisanalytics-streams.go
```diff
@@ -0,0 +1,68 @@
+package resources
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/kinesisanalytics"
+)
+
+type KinesisAnalyticsStream struct {
```
bethge: Why do you call it `KinesisAnalyticsStream` instead of `KinesisAnalyticsApplication`?

resources/kinesisanalytics-streams.go
```diff
@@ -0,0 +1,68 @@
...
+
+		params.ExclusiveStartApplicationName = lastApplicationName
+	}
+
+	return resources, nil
+}
+
+func (f *KinesisAnalyticsStream) Remove() error {
+
+	output, _ := f.svc.DescribeApplication(&kinesisanalytics.DescribeApplicationInput{
```
bethge: This can easily fail. It's safer to check if an error occurred and return it in that case with:
```go
output, err := ...
if err != nil {
    return err
}
...
```

resources/kinesisanalytics-streams.go
tomvachon: Because I was apparently tired, Ill fix it now

tomvachon:
PR updated and squashed

PR #121, opened 2018-03-08
Add CloudHSMv2 support

tomvachon:


resources/cloudhsmv2-clusterhsms.go
```diff
@@ -0,0 +1,72 @@
...
+
+	for {
+		resp, err := svc.DescribeClusters(params)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, cluster := range resp.Clusters {
+			var hsmIds []*string
+			for _, hsm := range cluster.Hsms {
```
bethge: Opposed to other resources, here you delete all HSMs of a given cluster. Although by itself not an issue, the problem arises when wanting to protect resources from deletion. For this we use the `String()` function, used here: https://github.com/rebuy-de/aws-nuke/blob/5002410cbea40bf9bc007521d0c1a72b3f107eb4/cmd/nuke.go#L176:L180

Instead of a struct of `CloudHSMV2ClusterHSM` with a slice of `hsmIDs` a single `hsmID` for each `CloudHSMV2ClusterHSM` would probably make sense. This would simplify the `Remove()` method and if the `String()` would return the `hsmID` we can protect a single HSM resource from deletion.

I might be overlooking something though.

tomvachon: Yea, I was starting to think that way this morning anyways.

tomvachon:
This has been updated to work on the HSM's one at a time, which is working better actually

PR #112, opened 2018-03-06
Feature/devicefarm

tomvachon:
How should I handle Regions which dont support the service (or make the error cleaner)

ERRO[0001] Listing with resources.ResourceLister failed. Please report this to https://github.com/rebuy-de/aws-nuke/issues/new.
    !!! RequestError: send request failed
    !!! caused by: Post https://devicefarm.us-east-1.amazonaws.com/: dial tcp: lookup devicefarm.us-east-1.amazonaws.com: no such host
ERRO[0002] Listing with resources.ResourceLister failed. Please report this to https://github.com/rebuy-de/aws-nuke/issues/new.
    !!! RequestError: send request failed
    !!! caused by: Post https://devicefarm.us-east-2.amazonaws.com/: dial tcp: lookup devicefarm.us-east-2.amazonaws.com: no such host
us-west-2 - DeviceFarmProject - 'arn:aws:devicefarm:us-west-2:999999999999:project:fc78472b-96d0-4234-aa2e-0cfa72dc3db6' - would remove
Scan complete: 1 total, 1 nukeable, 0 filtered.

Issue #63, opened 2017-12-12
DNS Error


sabrehagen:
[This is a run](https://gist.github.com/sabrehagen/d01fb2f2e05e99ad968839d0fdebf42c) with errors and private data redacted.

The DNS lookups fail at consistent intervals. I've included the full log with the errors inline such that you might be able to correlate the timing of the error with a certain part of the resource deletion sequence.

`aws-nuke` `v.1.3.0`

svenwltr:
Hello @sabrehagen,

apparently there is [no EFS](http://docs.aws.amazon.com/general/latest/gr/rande.html#elasticfilesystem-region) in `eu-west-1`. I don't know why the AWS SDK is giving such a confusing error message. We should try to handle that error better.

tomvachon:
This is a larger problem, It's coming up in #112 as well

svenwltr:
This is fixed in #124 and will be part of the next release.

PR #124, opened 2018-03-08
Proposed fix for some of the errors around new services

tomvachon:
This might help #112 and #63 

Its not the most "elegant" solution but considering we can't even get a response from AWS which says ServiceUnsupported or the like, it might be all we can.  The only alternative I can think of is making a mapping of all services to all endpoints which will get messy to maintain.

### DIFF
--- a/cmd/scan.go
+++ b/cmd/scan.go
@@ -3,9 +3,9 @@ package cmd
 import (
 	"fmt"
 	"runtime/debug"
-	"strings"
 
 	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/rebuy-de/aws-nuke/pkg/awsutil"
 	"github.com/rebuy-de/aws-nuke/pkg/util"
 	"github.com/rebuy-de/aws-nuke/resources"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


svenwltr:
The endpoints are already [maintained by the SDK](https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/). I added a validation check before the actual request get sent.

What do you think?

(I actually didn't want to push into your branch, but it happened)

tomvachon:
That seems reasonable, it may cause more updates in the long run of the SDK.  Does the v2 SDK do it as well?

svenwltr:
It does: https://godoc.org/github.com/aws/aws-sdk-go-v2/aws/endpoints

svenwltr:

svenwltr:
Unfortunately this doesn't cover all cases. For example Lighsail is available in all regions, but the Lighsail domain isn't:

```
ERRO[0004] Listing with LightsailDomain failed. Please report this to https://github.com/rebuy-de/aws-nuke/issues/new.
    !!! InvalidInputException: Domain-related APIs are only available in the us-east-1 Region. Please set your Region configuration to us-east-1 to create, view, or edit these resources.
    !!! 	status code: 400, request id: cf5a1f23-26a1-11e8-b93c-77bf2f94ebab
```

PR #134, opened 2018-03-11
add session token support

### LINKED ISSUE #126: support STS login
I use aws-vault to login to aws. 
as such there is no aws access keys, only sts temporary tokens.

using --profile Default works, but is not as intuitive.

svenwltr:
> Closes #126


svenwltr:
bump

PR #141, opened 2018-03-14
Add SSM Support

tomvachon:


resources/ssm-associations.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	return resources, nil
+}
+
+func (f *SSMAssociation) Remove() error {
+
+	_, err := f.svc.DeleteAssociation(&ssm.DeleteAssociationInput{
+		AssociationId: f.associationID,
+		//InstanceId:    f.instanceID,
```
bethge: ?

resources/ssm-patchbaselines.go
```diff
@@ -0,0 +1,69 @@
...
+			return nil, err
+		}
+
+		for _, baselineIdentity := range output.BaselineIdentities {
+			if baselineIdentity.BaselineDescription == nil {
+				resources = append(resources, &SSMPatchBaseline{
+					svc: svc,
+					ID:  baselineIdentity.BaselineId,
+				})
+			} else if !strings.Contains(*baselineIdentity.BaselineDescription, "Provided by AWS") {
```
bethge: To filter resources I would generally recommend the `Filter()` function. E.g.: https://github.com/rebuy-de/aws-nuke/blob/f9a3a639fe63fc9be7060556f53335635ce5da74/resources/iam-role-policy-attachments.go#L53-L58

In this case, I actually think your approach in `SSMDocument` is really great. It reduces the number of resources returned and let's AWS do the filter work. I tried it briefly with `aws ssm describe-patch-baselines --filters="Key=OWNER,Values=Self"` and it looks promising. Wdyt?

tomvachon: I have an issue with Patch Baselines, which I am not sure will be easy to resolve, you can't delete the default.  So either we accept that, or we make a ton of conditional coding to find the right AWS provided one which matches the right OS.  Thoughts?

tomvachon:
Moving this to a top level question:

I have an issue with Patch Baselines, which I am not sure will be easy to resolve, you can't delete the default. So either we accept that, or we make a ton of conditional coding to find the right AWS provided one which matches the right OS. Thoughts?

Its a bit irrelevant to this, but I wanted to mention it

resources/ssm-patchbaselines.go
bethge: Wouldn't filtering by `Owner=Self` resolve that issue?

tomvachon:
@bethge moving this to Top level, no, it doesnt.  You can make your own a default.  Then you need to make a non-nuked resource one, it might be a "put a pin in it" issue and see how often it creeps up

resources/ssm-associations.go
tomvachon: Whoops I was testing something. Let me double check the results but I dont think it was a mandatory flag

tomvachon:
I removed the comment after a test to ensure that was proper

bethge:
Ah ok, I see the problem now.

Looking at [Terraform's implementation of PatchBaseline][1], they also do not handle deleting the default PatchBaseline deletion. 

I would suggest we gracefully handle this case by filtering `PatchBaseline` resources with `DefaultBaseline=true` via the `Filter()` function. That would allow us to tell the user why we cannot delete that particular `PatchBaseline` resource.

[1]: https://github.com/terraform-providers/terraform-provider-aws/blob/e5bd586b0e44502aa898b77479fa6cb6eaa21e0c/aws/resource_aws_ssm_patch_baseline.go#L247-L262

tomvachon:
@bethge resolved

Issue #150, opened 2018-03-21
Add support for regex in resources filter


svenwltr:
### Problem

Currently the resource filter only supports exact string matches of the resource identifier (eg `s3://my-bucket/blubber.txt`). We should also support more flexible filter options. A first big step would be the support of regular expressions. Later we might want to add additional resource-specific filter methods which could rely on data which isn't part of the resource identifier (eg based on the Tag or AZ of an EC2 Instance; or access keys of a specific user).

### Proposal

My proposal for the configuration file is to optionally be able to specify a map instead of a string in the filter. This map should contain a `type` and optionally the data for the filter.

#### Example

Currently it look like this (which still should work in the future):

```yaml
accounts:
  "00000000000":
    filters:
      S3Bucket:
      - "s3://my-bucket/bish.txt"
      - "s3://my-bucket/bash.txt"
      - "s3://my-bucket/bosh.txt"
```

It should also be possible to write it like this:

```yaml
accounts:
  "00000000000":
    filters:
      S3Bucket:
      - type: exact
        id: "s3://my-bucket/bish.txt"
      - type: exact
        id: "s3://my-bucket/bash.txt"
      - type: exact
        id: "s3://my-bucket/bosh.txt"
```

Using a regex could look like this:

```yaml
accounts:
  "00000000000":
    filters:
      S3Bucket:
      - type: regex
        id: "s3://my-bucket/b[iao]sh\\.txt"
```

#### Alternatives

Alternatively the map could just contain a single `type` `value` pair, like this:

```yaml
accounts:
  "00000000000":
    filters:
      S3Bucket:
      - regex: "s3://my-bucket/b[iao]sh\\.txt"
```

But I think this would make things ambiguous. What if someone specified multiple types in the same filter? How should we handle extra data (in case of the extension to allow resource-specific filter)?

----

Please discuss.

bjoernhaeuser:
Probably most of the requirements could be fulfilled by simple glob matching, right? 

I would go for the first solution (`type: regex`).

svenwltr:
> Probably most of the requirements could be fulfilled by simple glob style matching, right?

Do you mean like this?

```yaml
accounts:
  "00000000000":
    filters:
      S3Bucket:
      - "s3://my-bucket/b*sh.txt"
```

This would work, but I do not see any guarantee that the wildcard `*` will never be part of an resource identifier.

And if you mean it like this:

```yaml
accounts:
  "00000000000":
    filters:
[...]

PR #165, opened 2018-03-22
Move config into separate package

svenwltr:
> In preparation of #150


pkg/config/config_test.go
```diff
@@ -44,16 +44,16 @@ func TestConfigBlacklist(t *testing.T) {
...
 
 func TestLoadExampleConfig(t *testing.T) {
-	config, err := LoadConfig("test-fixtures/example.yaml")
+	config, err := Load("test-fixtures/example.yaml")
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	expect := NukeConfig{
+	expect := Nuke{
```
stephanlindauer: without the "*Config" it feels like it's missing context, because the package it is in, is not something i'd look at first when guessing what a struct does. just my 2 cents...

svenwltr: This is convention: https://github.com/golang/go/wiki/CodeReviewComments#package-names

PR #166, opened 2018-03-22
Improve filters

### LINKED ISSUE #150: Add support for regex in resources filter

svenwltr:
> Fixes #150

I think this is the worst code I ever have written, but I do not see any better way with this complicated YAML library.

@rebuy-de/prp-aws-nuke Please tell me what you think before I continue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,36 +152,36 @@ Do you want to continue? Enter account alias to continue.
 > aws-nuke-example
 
 eu-west-1 - EC2DHCPOption - 'dopt-bf2ec3d8' - DependencyViolation: The dhcpOptions 'dopt-bf2ec3d8' has dependencies and cannot be deleted.
-	status code: 400, request id: 9665c066-6bb1-4643-9071-f03481f80d4e
+    status code: 400, request id: 9665c066-6bb1-4643-9071-f03481f80d4e
 eu-west-1 - EC2Instance - 'i-01b489457a60298dd' - triggered remove
 eu-west-1 - EC2KeyPair - 'test' - triggered remove
 eu-west-1 - EC2RouteTable - 'rtb-ffe91e99' - DependencyViolation: The routeTable 'rtb-ffe91e99' has dependencies and cannot be deleted.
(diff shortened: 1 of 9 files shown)


svenwltr:
I simplified the whole marshaling according to the talk with @bethge.

svenwltr:
@rebuy-de/prp-aws-nuke I now added a README and this PR is ready for review. So please review.

PR #167, opened 2018-03-23
Add AppStream 2.0 Service Coverage

tomvachon:
Note: per #122 I had to get a little creative around ImageBuilders.  I made a "Waiting" resource type which does nothing but ensure that whatever condition the ImageBuilder Remove() resource type needs can be satisified if its the last resource left in the Queue otherwise it will error out.

### DIFF
--- a/resources/appstream-directoryconfigs.go
+++ b/resources/appstream-directoryconfigs.go
@@ -1,0 +1,60 @@
+package resources
+
+import (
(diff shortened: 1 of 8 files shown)


tomvachon:
Can we get this into the 2.0.1 @svenwltr ?  I am not sure why its still hanging out

svenwltr:
> I am not sure why its still hanging out

Me neither :-)

PR #172, opened 2018-03-27
fix more typos - fixes #170

### LINKED ISSUE #170: resource-types exclude/include in config not working
adding a `include` or `exclude` filter via `resource-types` in the config file seems to have no effect.

resources are checked and removed anyway.

example config:

```
resource-types:
  include:
  - foo
  exclude:
  - S3Bucket
```

i've added a printf after the `yaml.Unmarshal` in `config.go` and the returned values are empty:

`ResourceTypes:{Targets:[] Excludes:[]}}] ResourceTypes:{Targets:[] Excludes:[]}`

I guess this is caused because `types.Collection` is used for these values and there seems to be no custom unmarshaler for this type.

lobeck:
it's always plural, see also #171

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,7 +257,7 @@ account-blacklist:
 
 resource-types:
   # don't nuke IAM users
-  exclude:
+  excludes:
   - IAMUser
 
 accounts:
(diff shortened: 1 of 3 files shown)


svenwltr:
Thank you very much!

PR #173, opened 2018-03-28
add fallback for missing services in endpoint list

### LINKED ISSUE #157: [BUG] Global Region Creates Race Conditions for Unknown Services
When we don't know a service's given endpoint, the current code determines it as both "Global" and n * regions.  As a result, the system tries to delete both at once, the problem arises when there is a implicit parent child dependancy.  

For example: Using MediaStoreData to delete each item in a MediaStore Container requires an endpoint which is unique for each container which are also regionally isolated.  When the "global" or "regional" delete occurs on the container, the listers end up continuing to try and iterate and it causes a Segfault for a nil value reference.
```
global - MediaStoreContainer - 'sadfdsgdsfgdsfgdsfg' - would remove
global - MediaStoreDataItems - 'ServicesToBlacklist.txt' - would remove
us-east-1 - MediaStoreContainer - 'sadfdsgdsfgdsfgdsfg' - would remove
us-east-1 - MediaStoreDataItems - 'ServicesToBlacklist.txt' - would remove
Scan complete: 4 total, 4 nukeable, 0 filtered.

Do you really want to nuke these resources on the account with the ID 99999999999 and the alias 'itsawsaccount-sandbox1'?
Do you want to continue? Enter account alias to continue.
> itsawsaccount-sandbox1

global - MediaStoreContainer - 'sadfdsgdsfgdsfgdsfg' - ContainerInUseException:
	status code: 400, request id: ES3S7MAXLS3OFWOSUEZH2VDDL24AZTO5NMY3MSM6TXC7SZHSQXF7NCFDFJSUPISXUIUGA53QUJETU2BSLFX7ILQ
global - MediaStoreDataItems - 'ServicesToBlacklist.txt' - triggered remove
us-east-1 - MediaStoreContainer - 'sadfdsgdsfgdsfgdsfg' - triggered remove
us-east-1 - MediaStoreDataItems - 'ServicesToBlacklist.txt' - ObjectNotFoundException: Object not found.
	status code: 404, request id: YNP7NR3DVES4G3GV2MI63GOMFAC5XLH3Y2SKWBKXOXKSGXWGDRFH62LW5U6PYL7NR6XCLL2YBLJKNF2IRMT4Y7A

Removal requested: 2 waiting, 2 failed, 0 skipped, 0 finished

global - MediaStoreContainer - 'sadfdsgdsfgdsfgdsfg' - ContainerInUseException:
	status code: 400, request id: 6LJILOE4ZER5BIP6GCPLVOJS5A377BKXRRP55KMGYYIGRAYEG6YHODVHMKJTYHQFZ5E6IHFNQW4X5BNW7CPAR7I
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x1e8a90e]

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x1e8a90e]

goroutine 1 [running]:
github.com/rebuy-de/aws-nuke/resources.ListMediaStoreDataItems(0xc4204cc480, 0xc4201f0000, 0x2718bfa, 0x13, 0xc4201fb758, 0x90)
	/Users/thv596/gopath/src/github.com/rebuy-de/aws-nuke/resources/mediastoredata-items.go:54 +0x56e
```

That code is here: https://github.com/tomvachon/aws-nuke/blob/feature/mediastore/resources/mediastoredata-items.go#L54

**My proposal**

If we cannot determine a endpoint for a given service, we assume its regional to prevent a race condition such as above.

svenwltr:
> fixes #157

@tomvachon Would that solve the problem? Could you verify it?

### DIFF
--- a/pkg/awsutil/session.go
+++ b/pkg/awsutil/session.go
@@ -142,7 +142,10 @@ func skipGlobalHandler(global bool) func(r *request.Request) {
 
 		rs, ok := endpoints.RegionsForService(endpoints.DefaultPartitions(), endpoints.AwsPartitionID, service)
 		if !ok {
-			// This means that the service does not exist and this shouldn't be handled here.
+			// This means that the service does not exist in the endpoints list.
+			if global {
+				r.Error = ErrSkipRequest(fmt.Sprintf("service '%s' is was not found in the endpoint list; assuming it is not global", service))
+			}
 			return
 		}
 


tomvachon:
Tested, looks great now

tomvachon:
Looks good, I had an additional logic check to add in MediaStore to prevent a race condition too

svenwltr:

PR #177, opened 2018-04-04
WIP: Initial Test Harness

bethge:
@rebuy-de/prp-aws-nuke WDYT?

This is rather a black box approach to the integration test. It heavily relies on Terraform. The resulting terraform state and configuration is not very pretty in the resources folder. I would look into moving it to `/tmp/`.

On the upside, the `resources/ssm-maintenancewindows_test.go` is rather concise.

More generally speaking, we would need an empty account to run these tests on.

It does take quite a while due to the 15s cooldown of `--force`. A flag or env variable to circumvent it would speed things up.

### DIFF
--- a/resources/main.tf
+++ b/resources/main.tf
@@ -1,0 +1,5 @@
+provider "aws" {
+  region = "eu-west-1"
+  version = "~> 1.11"
+}
+
--- a/resources/main_test.go
+++ b/resources/main_test.go
@@ -1,0 +1,121 @@
+package resources
+
+import (
+	"fmt"
+	"github.com/hashicorp/terraform/terraform"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"testing"
+)
+
+var debugFlag = true
+
+func TestMain(m *testing.M) {
+	// call flag.Parse() here if TestMain uses flags
+
+	// Setup
+	os.Setenv("TF_IN_AUTOMATION", "1")
+	if err := RunCmd("terraform", "init", "-input=false", "-no-color"); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	// Test
+	code := m.Run()
+
+	// Teardown
+	fmt.Println("Teardown: destroying all terraform resources of the test suite")
+	RunCmd("terraform", "destroy", "-input=false", "-no-color", "-auto-approve")
+
+	os.Exit(code)
+}
+
+func ResourceTypeTest(tfType, tfName, nukeType string, t *testing.T) error {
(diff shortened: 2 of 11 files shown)


resources/main_test.go
```diff
@@ -0,0 +1,132 @@
...
+	code := m.Run()
+
+	// Teardown
+	fmt.Println("Teardown: destroying all terraform resources of the test suite")
+	RunCmd("terraform", "destroy", "-input=false", "-no-color", "-auto-approve")
+
+	os.Exit(code)
+}
+
+func ResourceTypeTest(tfType, tfName, nukeType string, t *testing.T) error {
```
svenwltr: `*testing.T` should be the first parameter, because this is sort of a uber-object, which gets passed into all test helpers.

svenwltr: Since you have `*testing.T` you don't need to return an `error`. You could fail directly.

resources/main.tf
```diff
@@ -0,0 +1,5 @@
+provider "aws" {
```
svenwltr: I would call the file something like `provider.tf`.

resources/main_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"strings"
+	"testing"
+)
+
+var debugFlag = false
+
+func TestMain(m *testing.M) {
```
svenwltr: I think it is not necessary to use `TestMain` here, since we already have the helper function `ResourceTypeTest` and could do the set up and tear down there.

We might want to add a test which doesn't need Terraform into this package.

resources/main_test.go
```diff
@@ -0,0 +1,132 @@
+package resources
```
svenwltr: Maybe we should put the whole test helper code into its own package? This would prevent cluttering up the `resources` package.

resources/main_test.go
```diff
@@ -0,0 +1,132 @@
...
+		cmd.Stdout = os.Stdout
+	}
+	cmd.Stderr = os.Stderr
+	return cmd.Run()
+}
+
+func CreatePlan(targets []string) (*os.File, []string, error) {
+	var diffResources []string
+
+	f, err := ioutil.TempFile("/tmp", "AWSNuke")
```
svenwltr: You should delete that file again with `defer`.

bjoernhaeuser (commented):
I like. Especially the shortness of the test itself is very compelling to me:)

der-eismann:
Closed because nothing happened in almost 3 years. Anybody can feel free to give it a new try.

PR #191, opened 2018-05-17
add placement groups, fix security group circular deps, s3 bucket filtering enhancement

steved:
* add EC2 placement groups
* security groups: revoke all IP permissions before destroying

It's possible for security groups to have circular dependencies on each
other, preventing the actual removal from ever working. If we remove the
rules first, it's possible for a future retry to succeed.
* don't fail entire S3 bucket list on one bucket

It's possible for a single S3 bucket to be in an undesirable state (i.e.
unattached to a region) and this will cause all buckets to be filtered.

### DIFF
--- a/resources/ec2-placement-groups.go
+++ b/resources/ec2-placement-groups.go
@@ -1,0 +1,63 @@
+package resources
+
+import (
+	"fmt"
+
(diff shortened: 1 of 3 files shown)


steved:
cc @svenwltr @tomvachon

PR #198, opened 2018-06-15
 add advanced filters via properties

svenwltr:
@rebuy-de/prp-aws-nuke Please take a look. What do you think?

/cc @tomvachon

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ eu-west-1 - EC2Subnet - 'subnet-154d844e' - would remove
 eu-west-1 - EC2Volume - 'vol-0ddfb15461a00c3e2' - would remove
 eu-west-1 - EC2VPC - 'vpc-c6159fa1' - would remove
 eu-west-1 - IAMUserAccessKey - 'my-user -> ABCDEFGHIJKLMNOPQRST' - would remove
-eu-west-1 - IAMUserPolicyAttachment - 'my-user -> AdministratorAccess' - would remove
+eu-west-1 - IAMUserPolicyAttachment - 'my-user -> AdministratorAccess' - [UserName: "my-user", PolicyArn: "arn:aws:iam::aws:policy/AdministratorAccess", PolicyName: "AdministratorAccess"] - would remove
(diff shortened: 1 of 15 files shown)


tomvachon:
Can’t look for 4-5 days. I’ll get back to you

README.md
```diff
@@ -313,6 +313,29 @@ Any resource whose resource identifier exactly matches any of the filters in
...
+#### Filter Properties
+
+Some resources support filtering via properties. When a resource support these
+properties, they will be listed in the output like in this example:
+
+```
+global - IAMUserPolicyAttachment - 'admin -> AdministratorAccess' - [RoleName: "admin", PolicyArn: "arn:aws:iam::aws:policy/AdministratorAccess", PolicyName: "AdministratorAccess"] - would remove
+```
+
+To use properties, it is required to specify a object with `propeties` and
```
bjoernhaeuser: `properties`

cmd/queue.go
```diff
@@ -49,6 +52,20 @@ func (i *Item) List() ([]resources.Resource, error) {
...
 }
 
+func (i *Item) GetProperty(key string) string {
+	if key == "" {
+		return i.Resource.String()
+	}
+
+	getter, ok := i.Resource.(resources.ResourcePropertyGetter)
+	if !ok {
+		logrus.Warnf("%T does not support custom properties", i.Resource)
```
bjoernhaeuser: IMHO we should fail if the user provided config is incorrect

svenwltr: In this case I have to see how to bubble this error up, since we are in a goroutine. The only way that currently comes into my mind is an ugly error channel.

svenwltr: Will do in a separate PR to keep the history clean: https://github.com/rebuy-de/aws-nuke/pull/200

tomvachon:
As written, its hard to follow.  I dont see any major concerns tho

svenwltr:
I needed to do some additional changes to avoid nil pointer dereferences. See 6377b788d7a709518a5e756cda2b9685f2c371c2.

PR #217, opened 2018-06-26
make resource ID optional

svenwltr:
Make resource IDs optional (ie IDs from `String()` function). New resources should use the property feature instead.

Making legacy to avoid awkward resource IDs like this in `wafregional-ip-set-ips.go`:

```golang
return fmt.Sprintf("%s -> %s:%s", *r.ipSetid, *r.descriptor.Type, *r.descriptor.Value)
```

Describing [`waf.FieldToMatch`](https://godoc.org/github.com/aws/aws-sdk-go/service/waf#FieldToMatch) in a single string would be even more complex.

### DIFF
--- a/cmd/log.go
+++ b/cmd/log.go
@@ -27,8 +27,12 @@ func Log(region Region, resourceType string, r resources.Resource, c color.Color
 	fmt.Printf(" - ")
 	ColorResourceType.Print(resourceType)
 	fmt.Printf(" - ")
-	ColorResourceID.Printf("'%s'", r.String())
-	fmt.Printf(" - ")
+
+	rString, ok := r.(resources.LegacyStringer)
+	if ok {
+		ColorResourceID.Print(rString.String())
+		fmt.Printf(" - ")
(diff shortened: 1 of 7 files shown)


tomvachon:
Do we need to refactor for this? Seems like a big change.

I’d almost rather wait until the go sdk v2 since we will have to refactor
everything anyways

svenwltr:
@tomvachon We do not have to refactor every resource, now. Currently, both methods are possible and there is no need to touch the old resources. I would say we put this on the *aws-nuke* v3 roadmap.

tomvachon:
Agreed. Any clue how close v2 is for the sdk? I don’t track it closely

svenwltr:
> Any clue how close v2 is for the sdk? I don’t track it closely

No, I did not find any roadmap for v2.

svenwltr:
I had to update the PR, since the resource comparison did not work with the previous one. I am not very satisfied with this solution, but it works.

PR #223, opened 2018-06-28
Allow filtering EC2SecurityGroup by property "GroupName"

leflamm:
* also gitignoring .idea files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 /aws-nuke
 /aws-nuke-*
 .glide
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,3 @@
 /aws-nuke
 /aws-nuke-*
 .glide
-.idea
--- a/resources/ec2-security-groups.go
+++ b/resources/ec2-security-groups.go
@@ -81,6 +81,11 @@ func (sg *EC2SecurityGroup) Remove() error {
 	return nil
 }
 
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -2,3 +2,4 @@
 /aws-nuke
 /aws-nuke-*
 .glide
+.idea
```
svenwltr: We do not want to add any IDE specific ignores here. Every developer should use any IDE or text editor they want and requiring to update our `.gitignore` to reflect the use of all the quirky files some IDEs produce would be a cumbersome tasks.

You rather should [configure a global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore).

resources/ec2-security-groups.go
```diff
@@ -81,6 +81,11 @@ func (sg *EC2SecurityGroup) Remove() error {
 	return nil
 }
 
+func (sg *EC2SecurityGroup) Properties() Properties {
+	return NewProperties().
+		Set("GroupName", sg.name)
```
svenwltr: I suggest to just call it `Name`.

.gitignore
leflamm: Understood

resources/ec2-security-groups.go
leflamm: Will do

leflamm:
Renamed "GroupName" to "Name", removed Idea-specific entry in `.gitignore`.

PR #224, opened 2018-07-02
Add Support for Perl Compatible Regular Expressions for Inversions

leflamm:
Hi,

after using the great `aws-nuke` tool for a while I found out that it didn't have support for Perl-style regular expressions. This would be particularly helpful because using such regular expressions one can (not only) realize inverse matches, like with a zero-width negative lookahead assertion
```yml
accounts:
  <some account>: # stage12
    filters:
      CloudFormationStack:
        - type: pcre
          value: "^((?!foo).)*$"
```
which filters all `CloudFormationStack`s that do _not_ contain the string "foo" or - in other words - will only remove stacks containing the string "foo".

I figured replacing the existing functionality behind the `regex` filter wouldn't be a good idea because it might introduce subtle changes and sneakily break existing `aws-nuke` configurations. So I introduced a separate keyword: `pcre`.

Cheers, Christian


leflamm:
Hi! Would appreciate help to better understand what went wrong during the travis build.

svenwltr:
Thanks for your PR @leflamm!

Unfortunately, the library you chose uses C bindings. Supporting this would have an implication:

* We would not be able to build a statically linked binary, anymore.

This means the produced binary would be platform dependent. More specifically, it would be dependent on the C standard library (libc) in POSIX systems. I am not sure, if it is even dependent on the libc version. Also the binary would be dependent on a certain version of the prce library.

In short, we would lose the ability to provide plain binary releases and would need to make them distribution-specific (ie `dep` and `rpm`) packages.

Furthermore, I have no idea how this works for MacOS, but it would definitively stop working for Windows. Even, if Windows is not officially supported here, it should just work, currently.

leflamm:
@svenwltr I see, I believe I'll provide a much simpler PR with a specific missing function instead of solving the problem with such a complicated "solution".

PR #226, opened 2018-07-02
Allow to invert Filter Results

leflamm:
Hi!

`invert: true` can be added to any filter statement and would invert such filter result. 

Can be useful to not only protect resources from being nuked, but to also explicitly define resources to be removed by inverting a filter result (else used for resource protection).

Cheers, Christian

### DIFF
--- a/README.md
+++ b/README.md
@@ -373,6 +373,17 @@ Route53HostedZone:
   value: "*.rebuy.cloud."
 ```
 
+####  Inverting Filter Results
+
+Any filter result can be inverted by using `invert: true`, for example:
+```yaml
+CloudFormationStack:
+- property: Name
+  value: "foo"
+  invert: true
+```
+In this case *any* CloudFormationStack ***but*** the ones called "foo" will be filtered. Be aware that *aws-nuke* internally takes every resource and applies every filter on it. If a filter matches, it marks the node as filtered.
+
 ## Install
 
 ### Use Released Binaries
--- a/cmd/nuke.go
+++ b/cmd/nuke.go
@@ -189,6 +189,10 @@ func (n *Nuke) Filter(item *Item) error {
 			return err
 		}
 
+		if IsTrue(filter.Invert) {
+			match = !match
+		}
+
 		if match {
 			item.State = ItemStateFiltered
 			item.Reason = "filtered by config"
--- a/cmd/util.go
+++ b/cmd/util.go
@@ -38,3 +38,7 @@ func ResolveResourceTypes(base types.Collection, include, exclude []types.Collec
 
 	return base
 }
+
+func IsTrue(s string) bool {
+	return strings.TrimSpace(strings.ToLower(s)) == "true"
+}
--- a/cmd/util_test.go
+++ b/cmd/util_test.go
@@ -61,3 +61,19 @@ func TestResolveResourceTypes(t *testing.T) {
 		})
 	}
 }
+
+func TestIsTrue(t *testing.T) {
+	falseStrings := []string{"", "false", "treu", "foo"}
+	for _, fs := range falseStrings {
+		if IsTrue(fs) {
+			t.Fatalf("IsTrue falsely returned 'true' for: %s", fs)
+		}
+	}
+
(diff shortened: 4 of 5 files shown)


svenwltr:
Hello @leflamm,

what would happen, if you specify more than one filter with `invert`? It looks like it would filter everything then.

leflamm:
Hi @svenwltr, can you give a specific example? Thx!

leflamm:
Hi @svenwltr , I believe you are correct regarding two inverted exact matches.

IIUC, for `glob` or `regex` it's slightly different. As implemented at the moment filters defined like this
```yml
accounts:
  <some account>: # stage12
    filters:
      CloudFormationStack:
        - type: glob
          value: "*u*"
          invert: true
        - type: glob
          value: "*o*"
          invert: true
```
would filter (as in "protect from removal") all resources that:
- contain "u"
- contain an "o"
- contain neither 

or to say it the other way: it would only remove resources containing both - "o" and "u".

leflamm:
But thanks for your hint! I believe I may have to think a little more about the desired behavior.

svenwltr:
*aws-nuke* internally takes every resource and applies every filter on it. If a filter matches, it marks the node as filtered.

So for example, assume we have these Cloud Formation Stacks with your example filter: `blub`, `blob`, `blib`, `bloub`

* `blub`: rule 1: false, rule 2: true → filtered
* `blob`: rule 1: true, rule 2: false → filtered
* `blib`: rule 1: true, rule 2: true → filtered
* `bloub`: rule 1: false, rule 2: false → not filtered

So you are right. The inverting thing confused me a bit. Is this the behavior you want?

Also /cc @rebuy-de/prp-aws-nuke

leflamm:
It makes perfect sense for `aws-nuke` to work that way and to apply filters that way. For such a tiny feature I would never interfere with that logic and change any of it.

It's definitely a behavior I can live with - users simply need to understand how filters get applied.

There may be users who write down this...
```yml
accounts:
  <some account>: # stage12
    filters:
      CloudFormationStack:
        - type: glob
          value: "*u*"
          invert: true
        - type: glob
          value: "*o*"
          invert: true
```
... and wrongly assume any CloudFormationStack containing either "u" or "o" or both to be marked for removal - which is not how applying filters works. Still, it can easily be achieved by e. g.:
```yml
accounts:
  <some account>: # stage12
    filters:
      CloudFormationStack:
        - type: regex
          value: ".*[uv].*"
          invert: true
```
Maybe the `README.md` could use a hint like
> aws-nuke internally takes every resource and applies every filter on it. If a filter matches, it marks the node as filtered.

Cheers, Christian

svenwltr:
Sorry for the late response. It feels like this is something we might want to implement in a different way (eg something not in `filter`). I need some days to think about this.

/cc @rebuy-de/prp-aws-nuke @tomvachon Maybe you have some opinions on this?

svenwltr:
Okay, since there are no other opinions and I still do not have new thoughts I guess we can just do this. Could you please improve the README like you suggested and squash your commits?

bjoernhaeuser:
@svenwltr Hm, i think that could easily be a foot-gun, right?

svenwltr:
@bjoernhaeuser In what sense? People doing the configuration wrong? Encouraging people to use this for a wrong use case?

bjoernhaeuser:
Yeah, or having a configuration which noone understands anymore if you need a lot of filtering. But on the other side you probably do not need more than one filter a lot of times..

bjoernhaeuser:
Therefore: 👍

leflamm:
> But on the other side you probably do not need more than one filter a lot of times

That's what I figured. Even if I want to nuke specific resources using 3 different patterns I can write
```yml
      <resource>:
        - type: regex
          property: <property>
          value: (<pattern1>|<pattern2>|<pattern3>)
          invert: true

```
I'll provide a README.md proposal and will squash the commits.

leflamm:
@svenwltr plz let me know if the README.md change works for you or if this needs to be more elaborate.

Cheers

Issue #228, opened 2018-07-04
Seems like not all SNSSubscriptions are found


leflamm:
Hi,

I'm using a simple config file to list all `SNSSubscriptions`s
```yml
---
regions:
  - global
  - eu-central-1

account-blacklist:
  - <some prod account> # prod

accounts:
  <some stage>: # stage04
    filters:

resource-types:
  targets:
    - SNSSubscription
```
When running `aws-nuke` with that config file it finds 18 SNSSubscriptions
```
aws-nuke version v2.1.0.13.g2d766a2 - Mi 4. Jul 16:45:03 CEST 2018 - 2d766a2056b00596c52ef3d199ec550d72ba8c9e

Do you really want to nuke the account with the ID 718925568337 and the alias 'stage04'?
Do you want to continue? Enter account alias to continue.
> stage04

eu-central-1 - SNSSubscription - Owner: <account> ARN: arn:aws:sns:eu-central-1:<account>:<some subscription id> - would remove
eu-central-1 - SNSSubscription - Owner: <account> ARN: arn:aws:sns:eu-central-1:<account>:<some subscription id> - would remove
eu-central-1 - SNSSubscription - Owner: <account> ARN: arn:aws:sns:eu-central-1:<account>:<some subscription id> - would remove
eu-central-1 - SNSSubscription - Owner: <account> ARN: arn:aws:sns:eu-central-1:<account>:<some subscription id> - would remove
eu-central-1 - SNSSubscription - Owner: <account> ARN: arn:aws:sns:eu-central-1:<account>:<some subscription id> - would remove
eu-central-1 - SNSSubscription - Owner: <account> ARN: arn:aws:sns:eu-central-1:<account>:<some subscription id> - would remove
eu-central-1 - SNSSubscription - Owner: <account> ARN: arn:aws:sns:eu-central-1:<account>:<some subscription id> - would remove
eu-central-1 - SNSSubscription - Owner: <account> ARN: arn:aws:sns:eu-central-1:<account>:<some subscription id> - would remove
eu-central-1 - SNSSubscription - Owner: <account> ARN: arn:aws:sns:eu-central-1:<account>:<some subscription id> - would remove
eu-central-1 - SNSSubscription - Owner: <account> ARN: arn:aws:sns:eu-central-1:<account>:<some subscription id> - would remove
eu-central-1 - SNSSubscription - Owner: <account> ARN: arn:aws:sns:eu-central-1:<account>:<some subscription id> - would remove
[...]

PR #229, opened 2018-07-04
#228: add paging support to SNSSubscription

leflamm:
* Add paging support to SNSSubscription
* supposed to fix https://github.com/rebuy-de/aws-nuke/issues/228


svenwltr:
Thank you!

PR #239, opened 2018-08-05
S3Bucket filtering with tags

nisabek:
It does add to the timing though, fetching all those tags for every bucket...

### DIFF
--- a/resources/s3-buckets.go
+++ b/resources/s3-buckets.go
@@ -24,9 +24,12 @@ func ListS3Buckets(s *session.Session) ([]Resource, error) {
 
 	resources := make([]Resource, 0)
 	for _, name := range buckets {
+		tags, _ := retrieveTags(svc, name)
+
 		resources = append(resources, &S3Bucket{
 			svc:  svc,
 			name: name,
+			tags: tags,
 		})
 	}
 
@@ -61,6 +64,7 @@ func DescribeS3Buckets(svc *s3.S3) ([]string, error) {
 type S3Bucket struct {
 	svc  *s3.S3
 	name string
+	tags []*s3.Tag
 }
 
 func (e *S3Bucket) Remove() error {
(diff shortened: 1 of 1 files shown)


svenwltr:
> It does add to the timing though, fetching all those tags for every bucket...

This only matters in a scenario, where one has a lot of buckets, right?

resources/s3-buckets.go
```diff
@@ -74,6 +76,35 @@ func (e *S3Bucket) Remove() error {
...
+	}
+
+	return result.TagSet, nil
+}
+
+func (e *S3Bucket) Properties() types.Properties {
+	properties := types.NewProperties()
+	properties.Set("Name", e.name)
+
+	tags, err := e.retrieveTags()
```
svenwltr: The `Properties()` function gets called quite often during the whole process. Therefore I suggest to move the `retrieveTags` to `DescribeS3Buckets`.

resources/s3-buckets.go
```diff
@@ -22,9 +24,12 @@ func ListS3Buckets(s *session.Session) ([]Resource, error) {
 
 	resources := make([]Resource, 0)
 	for _, name := range buckets {
+		tags, _ := retrieveTags(svc, name)
```
nisabek: I'm ignoring the error, as the bucket might simply not be tagged and then we get a lot of 404s. I could also check what kind of error it is, but returning empty tag list instead of nil seems to work

nisabek:
I moved it to ListS3Bucket instead, as the Describe is used from several other places, I'm not sure it's worth changing the signature to return S3Buckets instead of strings.
and yes, the performance is less affected, since I moved it out of Properties.

resources/s3-buckets.go
svenwltr: I think we should check the error, because if someone filters based on Tags and the Tags are missing it might delete a resource which is supposed get filtered.

nisabek: Not sure what to do if there is an error. Fail everything just because some bucket is not tagged? Skip the bucket just because it's not tagged?...

svenwltr: I see at least two possible errors here: 1: There are no tags. 2: There is a network issue.

Especially the second one is interesting, because if it is some kind of network hiccup the resource does not get the tags in the properties and *aws-nuke* would filter it, besides it has tags. In this case it would get deleted which would be very unfortunate for an user.

der-eismann:
I'm closing this since it was done in #383

Issue #240, opened 2018-08-08
Enable Paging for Cloudformation Stacks


leflamm:
Hi! Support is lacking at the moment. Only the first 100 CFN stacks will be found. Will provide a PR ASAP. Cheers!

leflamm:
https://github.com/rebuy-de/aws-nuke/pull/241

PR #241, opened 2018-08-08
#240: Enable paging for CFN stacks

leflamm:
* enables paging support for Cloudformation Stacks
* more than just first 100 stacks will be found


svenwltr:
Thank you very much!

Issue #238, opened 2018-08-01
Nuke CloudFormation stacks first


leozhad:
I would like to suggest that aws-nuke should nuke any existing CloudFormation stacks before anything else, when running on an account. This way, they will terminate any resource launched by those stacks, and cut down on wait times that would occur on if you terminate other resources first.

svenwltr:
I am not sure how this actually would improve the wait times. At least not in a way that would justify the increasing code complexity.

Currently `aws-nuke` does an initial scan. While doing this it does not remove anything, so we would not speed up anything there. After this is sends a delete request to each found resource, but does not actually wait for the resource to be gone. Actually, this depends on the resource and if the remove request is sync or async, but AFAIK the most ones are async. Therefore the maximum amount of time we could save, is the time of the first run.

Are there actually so many resources in your scenario, that the first run takes so long? Am I missing something else?

Any other opinions @rebuy-de/prp-aws-nuke?

leozhad:
I am not sure it would make a drastic difference -- but it seems logical to terminate CloudFormation stacks first, since this would in turn terminate any resources they launched cleanly.

srij91:
Only issue I came across is, if awsnuke deletes the resources first even before the CFN stack is deleted, the CFN stack deletion fails saying it is unable to find the resource to delete and gets stuck in _**Deletion failed**_ status. This is keeping awsnuke to go in loop and retry deleting CFN stack over and over. The moment I manually deleted CFN stack nuke job completed.

svenwltr:
> but it seems logical to terminate CloudFormation stacks first

Logical yes, but I am concerned about the increased complexity. Also the code base is currently very messy and adding more stuff would make it worse.

> Only issue I came across is, if awsnuke deletes the resources first even before the CFN stack is deleted, the CFN stack deletion fails saying it is unable to find the resource to delete and gets stuck in Deletion failed status. This is keeping awsnuke to go in loop and retry deleting CFN stack over and over. The moment I manually deleted CFN stack nuke job completed.

That sounds like a bug to me. If you are able to delete it manually, *aws-nuke* should also be able to delete it. We should take a look into this.

leflamm:
I usually delete CFN stacks first, too.

Reason being: I already ran into the situation that the role that was used to create the CFN stack was nuked before the stack. IIUC the role used to create a stack is used to destroy it as well.

[...]

PR #242, opened 2018-08-15
Retain all CloudFormation resources in Deletion of CloudFormationStack

stephanlindauer:
This would be my suggestion on how to deal with CloudFormationStacks that get stuck in _**Deletion failed**_ state.
By retaining all resources that belong to the CloudFormationStacks we delete **just** the CloudFormationStack resource and **don't** cause cascading deletions. 
IMHO this should be the desired behavior. As with other resources, we just try to delete the resource itsself **EXcluding** the onces that depend on it.
The resources that were created from the CloudFormationStack and are left over after the CloudFormationStack is gone, should be deleted by their own resource implementation.
What do you think? 

ref: https://github.com/rebuy-de/aws-nuke/issues/238
ping @rebuy-de/prp-aws-nuke

### DIFF
--- a/resources/cloudformation-stack.go
+++ b/resources/cloudformation-stack.go
@@ -44,9 +44,20 @@ type CloudFormationStack struct {
 }
 
 func (cfs *CloudFormationStack) Remove() error {
-	_, err := cfs.svc.DeleteStack(&cloudformation.DeleteStackInput{
+	retainableResources, err := cfs.svc.ListStackResources(&cloudformation.ListStackResourcesInput{
 		StackName: cfs.stack.StackName,
 	})
+
+	retain := make([]*string, 0)
+	for _, r := range retainableResources.StackResourceSummaries {
+		retain = append(retain, r.LogicalResourceId)
+	}
+
+	cfs.svc.DeleteStack(&cloudformation.DeleteStackInput{
+		StackName:       cfs.stack.StackName,
+		RetainResources: retain,
+	})
+
 	return err
 }
 


bjoernhaeuser:
The job of aws-nuke is to delete resources. Retaining them to delete them makes no sense to me. CloudFormation is a special snowflakes on aws, because it takes care of deleting the created resource for you. Why not leverage this to make it easier?

My proposal is to delete stacks first, wait for them to be deleted, and afterwards delete the rest of the config.

stephanlindauer:
@bjoernhaeuser 

> The job of aws-nuke is to delete resources

excactly. but it would do it in either case. in theory it shouldn't make a difference whether we utilize the cloudformation deletion feature or whether we delete them from their own resource implementation. the result is the same: stuff is deleted.
the "deleting resources outside of CFS"-approach would enable us on top to use the config to protect resources that would otherwise be deleted.

> CloudFormation is a special snowflakes on aws, because it takes care of deleting the created resource for you. Why not leverage this to make it easier?

because this has turned out to be tricky in practice:
- we would have to squeeze this feature into aws-nuke to have this one exception
- if a resource depends on something that was created via CFS then the deletion will fail (i can give you some examples from our own infrastructure where this would bite us, later in person)
- if the deletion fails the resource in CFS that caused it might get deleted by plain old aws-nuke resource deletion. so the next time we come around trying to delete the stack it will fail again because it tries to delete something that isn't there anymore. et voila: endless faillure loop. 

TL:DR: i do agree that deleting CFS without utilizing the deletion feature feels a bit odd, but weight against it's alternatives, i think it's what would make most sense for our purpose (and i think of the users of this tool as well).

PR #245, opened 2018-08-21
CFS deletion logic fix

stephanlindauer:
So basically we try a normal deletion and **always** return an error. In the next run the resource will either be deleted and we're happy and move on or it will be in `DELETION_FAILED` state in which case we delete it while retaining all sub-resources.

@rebuy-de/prp-aws-nuke

### DIFF
--- a/resources/cloudformation-stack.go
+++ b/resources/cloudformation-stack.go
@@ -46,16 +46,24 @@ type CloudFormationStack struct {
 }
 
 func (cfs *CloudFormationStack) Remove() error {
-	if *cfs.stack.StackStatus != cloudformation.StackStatusDeleteFailed {
+	o, err := cfs.svc.DescribeStacks(&cloudformation.DescribeStacksInput{
+		StackName: cfs.stack.StackName,
+	})
+	stack := o.Stacks[0]
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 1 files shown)


resources/cloudformation-stack.go
```diff
@@ -44,9 +46,42 @@ type CloudFormationStack struct {
 }
 
 func (cfs *CloudFormationStack) Remove() error {
-	_, err := cfs.svc.DeleteStack(&cloudformation.DeleteStackInput{
+	stack, err := cfs.svc.DescribeStacks(&cloudformation.DescribeStacksInput{
```
svenwltr: There is already an up-to-date object in `cfs.stack`.

stephanlindauer: true. done. 
i also restructured the code a bit.

stephanlindauer: @svenwltr we actually need to get a fresh version of the stack object at this place. otherwise it will run into a race-condition where `DELETION_FAILED` is not set while it actually should be.

svenwltr: Ok, *aws-nuke* runs the re-scan after it tried to remove the resource. This means it gets refreshed after `Remove()` was called.

PR #250, opened 2018-08-28
compress release files

### LINKED ISSUE #249: Compress release assets
Hi,

I was starting to work on creating a Homebrew formula to fix #125, and noticed that the release binaries are not compressed. 

Would it be possible to add compressed versions of the binaries as well? This would allow brew to download the binary more quickly which on slow connections would be very beneficial IMO.

The file is almost 10x smaller when gzipped:

```
❯ ls -lh aws-nuke*
-rw-r--r--@ 1 mark  wheel   103M Aug 20 05:54 aws-nuke
-rw-r--r--  1 mark  wheel    16M Aug 28 14:14 aws-nuke.tar.gz
```

svenwltr:
> Fixes #249

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /vendor
 /aws-nuke
 /aws-nuke-*
+/releases
 .glide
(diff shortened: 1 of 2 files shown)


svenwltr:

PR #259, opened 2018-09-06
Add --max-wait-retries flag to bail out early if the script is stuck waiting

scottybrisbane:
We want to get to a point where we can run `aws-nuke` through our CI/CD platform and have noticed that it can get stuck in a waiting state for a considerable length of time in some scenarios. 

This pull request aims to resolve that by adding a flag that, if set, will cause the program to bail out if it is waiting for `n` number of iterations and no new work has been discovered in that time.

### DIFF
--- a/cmd/nuke.go
+++ b/cmd/nuke.go
@@ -84,6 +84,7 @@ func (n *Nuke) Run() error {
 	}
 
 	failCount := 0
+	waitingCount := 0
 
 	for {
 		n.HandleQueue()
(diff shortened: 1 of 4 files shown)


svenwltr:
Hello @scottybrisbane,

can you elaborate in what cases a resource stucks in `Waiting` state? This sounds like there is a bug in one of our resources.

scottybrisbane:
Hi @svenwltr, the primary case in which something was truly 'stuck' was a CloudFormation stack that had gotten into a nasty state with all resources it managed having already been deleted. It got confused by this and refused to delete, causing aws-nuke to keep that resource in a waiting state. The other case was in deleting an ElasticSearch Domain, which can take a considerable amount of time. It wasn't really 'stuck' as such, but would have left aws-nuke in a waiting loop for a considerable amount of time whilst it was being deleted.

I haven't experienced it myself, but a colleague mentioned that they also had an issue with an OpsWorks stack in which aws-nuke entered a loop as it couldn't properly delete the resources.

svenwltr:
I see. We already fixed the CloudFormation stack.

I think we could merge this PR, but it is essentially just a workaround for bugs.

scottybrisbane:
Thanks for the reviews!

@svenwltr I do agree that it's essentially just a workaround for bugs, but it should better enable us (and others) to run the tool from our CI/CD platform, without fear of the program entering a long-running loop.

PR #272, opened 2018-09-28
Empty objects from S3 bucket before nuking

### LINKED ISSUE #263: S3Bucket nuking doesn't function if the bucket is not empty
I have been contemplating submitting a PR to modify the logic for S3Bucket nuking to go through and delete all the objects in a bucket before trying to remove the bucket. This is due to the fact that nuking a non-empty bucket doesn't work, and in my opinion it should as the tool is aimed at being a very destructive cleanup tool.

We have performance issues (including OOM) with nuking S3Objects, due to the fact that it lists every single object in all buckets and stores this in memory (see #262). Ideally, we don't actually need to nuke S3Objects since we should be able to just nuke the entire bucket. Due to the above issue, however, we need to nuke S3Objects as well as S3Buckets.

Would the maintainers be open to such a PR?

scottybrisbane:
As per https://github.com/rebuy-de/aws-nuke/issues/263, deletes all objects from a bucket before attempting to delete the bucket itself.

Closes #263

### DIFF
--- a/resources/s3-buckets.go
+++ b/resources/s3-buckets.go
@@ -3,9 +3,11 @@ package resources
 import (
 	"fmt"
 
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


scottybrisbane:
There is one caveat to this approach. If you are nuking both `S3Bucket`'s and `S3Object`'s, the deletion of the bucket will cause 404's for the deletion of the objects in the first pass:
```
ap-southeast-2 - S3Object - s3://sbrisbane-test/test-1538173738 - [Bucket: "sbrisbane-test", Key: "test-1538173738", VersionID: "null", IsLatest: "true"] - NoSuchBucket: The specified bucket does not exist
	status code: 404, request id: 0B00AA0C281B526F, host id: GZh5rmLMMRzuIqWPJIt0yPjhOijOTRaqxgxrnBWNQ7T3D1MXBY8+aTSpBcF3BGR9IdS9KBkDhAc=
```
This is resolved in the next pass however, with the objects reported as removed:
```
ap-southeast-2 - S3Object - s3://sbrisbane-test/test-1538173738 - [Key: "test-1538173738", VersionID: "null", IsLatest: "true", Bucket: "sbrisbane-test"] - removed
```
Personally, I think this is acceptable as the overall status of the nuke is still successful:
```
Nuke complete: 0 failed, 2 skipped, 170 finished.
```
Felt it worth raising however.

svenwltr:

svenwltr:
> Personally, I think this is acceptable as the overall status of the nuke is still successful:

@scottybrisbane Yes it is. I think this also happens with other resources.

PR #273, opened 2018-09-29
Add ElasticacheReplicationGroup resource

scottybrisbane:
Adds an `ElasticacheReplicationGroup` resource. There is some overlap between this resource and the `ElasticacheCacheCluster` resource in that a replication group consists of multiple cache clusters. The problem though is that cache clusters that are part of a replication group aren't being deleted on their own, due to errors similar to the following:
```
InvalidCacheClusterState: Cache cluster transactions-001 is serving as primary for replication group transactions and cannot be deleted. To delete the entire replication group, use DeleteReplicationGroup.
--
  | status code: 400, request id: eb3f1c2e-c23f-11e8-a902-65beaf280fb2
```

### DIFF
--- a/resources/elasticache-replicationgroups.go
+++ b/resources/elasticache-replicationgroups.go
@@ -1,0 +1,54 @@
+package resources
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/elasticache"
+)
+
+type ElasticacheReplicationGroup struct {
+	svc     *elasticache.ElastiCache
+	groupID *string
+}
+
+func init() {
+	register("ElasticacheReplicationGroup", ListElasticacheReplicationGroups)
+}
+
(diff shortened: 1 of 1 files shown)


resources/elasticache-replicationgroups.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func init() {
+	register("ElasticacheReplicationGroup", ListElasticacheReplicationGroups)
+}
+
+func ListElasticacheReplicationGroups(sess *session.Session) ([]Resource, error) {
+	svc := elasticache.New(sess)
+
+	params := &elasticache.DescribeReplicationGroupsInput{MaxRecords: aws.Int64(100)}
```
svenwltr: We should check if there are additional pages. See https://github.com/rebuy-de/aws-nuke/blob/592a342a1848d062ee4fd6118e0089a138c7d2af/resources/databasemigrationservice-certificates.go#L39-L43 for an example.

scottybrisbane: @svenwltr apologies for the incredibly delayed response. I've added support for pagination. Would you mind reviewing this again?

PR #282, opened 2018-10-19
Add tags to more resources

mwhittington21:
I've mostly copied what was done over in #237 and extended it to resource types that I need to filter based on tag values.

There were two resource types that I've made a potentially controversial decision about: `EC2InternetGatewayAttachment`, `EC2VPNGatewayAttachment`, and `EC2VPCEndpoint`. For `EC2InternetGatewayAttachment` and `EC2VPNGatewayAttachment` I have chosen to inherit the tags from the Internet/VPN Gateway as a gateway would be more likely to have more specific tags than the VPC it lives in. If we should include both sets of tags I'm also fine with that, I just can't think of a good way to display both with the current key/value pairing.

For EC2VPCEndpoint I've just inherited the tags from the containing VPC.

### DIFF
--- a/pkg/types/properties.go
+++ b/pkg/types/properties.go
@@ -58,19 +58,28 @@ func (p Properties) Set(key string, value interface{}) Properties {
 }
 
 func (p Properties) SetTag(tagKey *string, tagValue interface{}) Properties {
+	return p.SetTagWithPrefix("", tagKey, tagValue)
+}
+
+func (p Properties) SetTagWithPrefix(prefix string, tagKey *string, tagValue interface{}) Properties {
 	if tagKey == nil {
 		return p
 	}
 
-	keyStr := *tagKey
+	keyStr := strings.TrimSpace(*tagKey)
+	prefix = strings.TrimSpace(prefix)
 
-	sanitizedTagKey := strings.Trim(keyStr, " ")
-	if sanitizedTagKey == "" {
+	if keyStr == "" {
 		return p
 	}
 
-	key := "tag:" + keyStr
-	return p.Set(key, tagValue)
+	if prefix != "" {
+		keyStr = fmt.Sprintf("%s:%s", prefix, keyStr)
+	}
+
+	keyStr = fmt.Sprintf("tag:%s", keyStr)
+
+	return p.Set(keyStr, tagValue)
 }
 
 func (p Properties) Get(key string) string {
--- a/pkg/types/properties_test.go
+++ b/pkg/types/properties_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/rebuy-de/aws-nuke/pkg/types"
 )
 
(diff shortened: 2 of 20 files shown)


svenwltr:
> If we should include both sets of tags I'm also fine with that, I just can't think of a good way to display both with the current key/value pairing.

Yeah, I think from the user perspective this makes more sense. We would need to rethink the tag handling a bit. Currently we have `properties.SetTag("foo", "bar")`, which would result in `tag:foo = bar`. We could add another function like `properties.SetTagWithPrefix("vpc", "foo", "bar")`, which could result in something like `tag:vpc:foo = bar`. Does this make sense?

svenwltr:
> Should we just merge this PR as-is and I can add that as a follow up?

I would rather not do that, because once we merge it people might rely on this. This means someone uses this as a filter and after we merge this, the filter would not be valid anymore and remove resources which actually should get filtered.

> I'll have a look at implementing that. 

If you want I can add this suggested function for you.

mwhittington21:
Sure @svenwltr if you want to add that and replace the functionality on those two resources that would be great.

svenwltr:
@mwhittington21 I added them.

mwhittington21:
That looks great, thanks a lot.

mwhittington21:
Any idea when this might get merged / a new version released? It's not urgent as I can build from source until then, I'm just curious.

svenwltr:
Sorry for that delay.

iainelder:
New aws-nuke user here. I have a suggestion based on my first impressions of the tool.

I found this issue while trying to figure out how to exclude the default networking setup of a new AWS account from being nuked. This includes retaining the default internet gateway and its attachment to the default VPC.

The default gateway's lack of a `Name` tag allows it to be filtered like this:

```yaml
      EC2InternetGateway:
        - property: tag:Name
          value: ""
```

But how to filter the default attachment? I don't think I would have figured it out without this comment about inheritance:

> For EC2InternetGatewayAttachment and EC2VPNGatewayAttachment I have chosen to inherit the tags from the Internet/VPN Gateway as a gateway would be more likely to have more specific tags than the VPC it lives in.

And so, with inheritance, the default attachment can be filtered like this:

```yaml
      EC2InternetGatewayAttachment:
        - property: tag:Name
          value: ""
```

Is this schema documented somewhere? Could it be made discoverable from from the output? Currently an EC2InternetGatewayAttachment shows `[]` where normally there should be filterable properties.

> If we should include both sets of tags I'm also fine with that, I just can't think of a good way to display both with the current key/value pairing.

How about doing it like in the IAMRolePolicy, where the properties from the role are prefixed with `role:`? For example, you can filter a role policy by the role name with this syntax:

```text
      IAMRolePolicy:
        - property: role:RoleName
          value: OrganizationAccountAccessRole
```

In this case we could have `vpc:` and `igw:` prefixes.

For my use case, it's actually more interesting to inherit the tags from the VPC. Based on what I knew about IAM role policies, here is what I tried before finding this issue:

```yaml
      EC2InternetGatewayAttachment:
        - property: vpc:IsDefault
          value: "true"
```

svenwltr:
Hello. This is a very old PR and does not seem to be directly related to you question, but I'll answer anyways.

> The default gateway's lack of a Name tag allows it to be filtered like this: `[...]`

Well this might work in most cases, but not in every one since it is possible to create IGWs, that do not have a name and therefore not a `Name` tag. Also it seems to be fine to add a `Name` tag to default VPC resources.

> Is this schema documented somewhere? Could it be made discoverable from from the output? Currently an EC2InternetGatewayAttachment shows [] where normally there should be filterable properties.

Yes it should be discoverable from the output, since it lists all properties. Seeing `[]` means either that there are no tags or there is a bug. The inherited tags would look like `tag:igw:Name` and `tag:vpc:Name`.

> For my use case, it's actually more interesting to inherit the tags from the VPC.

For correctness: You are talking about (aws-nuke) properties and not (AWS) tags. AWS Tags are embedded in the properties of some resources and are prefixed with `tag:`. There can be more properties that are defined by the code of each individual resource.

Nevertheless, inheriting the properties of the VPC for the InternetGatewayAttachment is not that easy with the current application design, because the resources are not connected internally, which mean we would have to fetch the IGW and then the actual VPC for each IGW Attachment.

iainelder:
@svenwltr, thanks the info. You raise a valid point about the mutability of tags that actually has helped me to think of a simpler solution (just save in the config the IDs of the default network resources for each account).

I don't want to make any more unrelated noise here so if I have more questions I'll create a new issue. Thanks for your patience :-)

PR #284, opened 2018-10-21
New Resources: IAMUsersMFA, SSMPatchBaselineGroup and IoT associations and fixes

jakubincloud:
Hi,

We are using aws-nuke for our Cloud vending machine at Cloudreach. This allows us to test aws-nuke with various resources. The result of that are new resources as:

IAMUserMFAattachment  
SSMPatchBaselineGroup
IoTCertificate - ( deactivate Certificate before deletion )
IoTPolicyPrincipalAttachment
IoTThingPrincipalAttachment

### DIFF
--- a/resources/iam-user-mfa-attachment.go
+++ b/resources/iam-user-mfa-attachment.go
@@ -1,0 +1,71 @@
+package resources
+
+import (
+	"fmt"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/rebuy-de/aws-nuke/pkg/types"
+)
+
+type IAMUserMfaAttachment struct {
+	svc          *iam.IAM
+	userName     string
+	serialNumber string
+}
+
+func init() {
+	register("IAMUserMfaAttachment", ListIAMUserMfaAttachments)
+}
+
+func ListIAMUserMfaAttachments(sess *session.Session) ([]Resource, error) {
+	svc := iam.New(sess)
+
+	resp, err := svc.ListUsers(nil)
+	if err != nil {
+		return nil, err
+	}
+
+	resources := make([]Resource, 0)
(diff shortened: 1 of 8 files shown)


jakubincloud:
The travis has problems with linting in golang 1.11
https://github.com/golang/go/issues/28291

jakubincloud:
The travis has problems with linting in golang 1.11
https://github.com/golang/go/issues/28291

jakubincloud:
Right now Github has issues with the storage layer so some comments / PR are not available.

the PR #285 (update golang from 1.8 to 1.11 )  makes the travis run successful.

resources/iam-user-mfa-attachment.go
```diff
@@ -0,0 +1,86 @@
...
+		if err != nil {
+			return nil, err
+		}
+		resources = append(resources, userMfaDevices...)
+	}
+
+	return resources, nil
+}
+
+func (f *IAMUser) ListAttachedMFADevices() ([]Resource, error) {
```
bjoernhaeuser: Is there no way to list all devices? This is in the end n+1 thingy here, and we are calling the API for every user once. Could slow down a lot.

svenwltr: That's true. I think the `ListMFADevice` just does not need to have a `UserName` argument. At least [the docs](https://docs.aws.amazon.com/sdk-for-go/api/service/iam/#ListMFADevicesInput) do not mention it and the equivalent CLI command `aws iam list-mfa-devices` does not need it.

jakubincloud: and I guess the same thing goes for ssh keys ( new thing in IAM that also prevents removing an user completely).
I just saw other comments and I will work on the proposed changes.

jakubincloud:
Please close this PR as I will submit new changes in a separate PRs. Thanks

PR #285, opened 2018-10-22
update golang 1.11

svenwltr:
We did not update for a while. Also we are getting errors from golint:

```
Step 6/14 : RUN go get -u golang.org/x/lint/golint
 ---> Running in 134e81a6fd1b
# golang.org/x/tools/go/internal/gcimporter
src/golang.org/x/tools/go/internal/gcimporter/bexport.go:212: obj.IsAlias undefined (type *types.TypeName has no field or method IsAlias)
The command '/bin/sh -c go get -u golang.org/x/lint/golint' returned a non-zero code: 2
```

See https://github.com/golang/go/issues/28291


svenwltr:
I also had to fix the Makefile. (Now the obligatory,) I have no idea how this ever worked.

PR #300, opened 2018-11-22
add lambda-event-source-mapping

### LINKED ISSUE #298: AWS-Nuke does not remove lambda event source maping
Hello!

The nuke does not remove any event source mapping that is attached to lambda when it runs causing conflicts when we redeploy after a nuke. I've had a scan through the resources and cannot find the functionality to remove it meaning we're having to use the CLI to do it after a nuke.
```
aws cli lambda delete-event-source-mapping -- --uuid UUID-HERE
```
Example command of removing with CLI

svenwltr:
> Fixes #298

/cc @preeny95

### DIFF
--- a/resources/lambda-event-source-mapping.go
+++ b/resources/lambda-event-source-mapping.go
@@ -1,0 +1,61 @@
+package resources
+
+import (
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/lambda"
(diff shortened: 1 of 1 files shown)


svenwltr:
The build fails, because I accidentally appended `.go` to the branch name and `gofmt` thinks a `.git` file is a source file:

```
 ./.git/logs/refs/heads/add-lambda-event-source-mapping.go:1:105: illegal character U+0040 '@'
```

I think making a new PR for that is not worth the hassle.

PR #304, opened 2018-11-27
New resource CognitoUserPoolDomain

sas-pemcne:
I was getting an error when trying to delete CognitoUserPools in our environment:

```
CognitoUserPool - foo - InvalidParameterException: User pool cannot be deleted. It has a domain configured that should be deleted first.
	status code: 400
```

This will describe each user pool, since that's the only way to grab the domain if it exists, and store that domain in the struct. On remove, if the domain is present, it removes that first. I tested this on my end with UserPools with and without a domain and it functions as expected.

### DIFF
--- a/resources/cognito-userpools.go
+++ b/resources/cognito-userpools.go
@@ -7,10 +7,9 @@ import (
 )
 
 type CognitoUserPool struct {
-	svc    *cognitoidentityprovider.CognitoIdentityProvider
-	name   *string
-	id     *string
-	domain *string
+	svc  *cognitoidentityprovider.CognitoIdentityProvider
+	name *string
+	id   *string
 }
 
 func init() {
@@ -32,23 +31,10 @@ func ListCognitoUserPools(sess *session.Session) ([]Resource, error) {
 		}
 
 		for _, pool := range output.UserPools {
-			describeParams := &cognitoidentityprovider.DescribeUserPoolInput{
-				UserPoolId: pool.Id,
-			}
-			userpool, err := svc.DescribeUserPool(describeParams)
-			if err != nil {
-				return nil, err
-			}
-			domain := ""
-			if userpool.UserPool.Domain != nil {
-				domain = *userpool.UserPool.Domain
-			}
-
 			resources = append(resources, &CognitoUserPool{
-				svc:    svc,
-				name:   pool.Name,
-				id:     pool.Id,
-				domain: &domain,
+				svc:  svc,
+				name: pool.Name,
+				id:   pool.Id,
 			})
 		}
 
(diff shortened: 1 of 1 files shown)


svenwltr:
Hello @sas-pemcne. To me it look like user pool domains are a separate source and should not be mixed with the user pool. What do you think?

sas-pemcne:
I thought that initially too but from what I've seen, the domains are unique (within the region) and tied directly to a UserPool. Additionally, the only way to get the domain information is by querying the `DescribeUserPool` api. The `DeleteUserPoolDomain` api also needs the UserPool id so that's why I put everything together here. 

If there was a separate way to query all domains then I would agree to throw it into a separate resource but it doesn't look like that's possible.

svenwltr:
Does it make sense to have a UserPool without a domain? If so, we should separate the resources, because someone might want to delete only the domains without the pools (for some reason). Since splitting them later would be a breaking change, I would prefer to do it now.

Since there is no list function, we could do it like in [DynamoDBTableItem](https://github.com/rebuy-de/aws-nuke/blob/master/resources/dynamodb-items.go) and reuse the list function.

What do you think?

sas-pemcne:
Ah, didn't think of that use case, that does make sense. I'll look into splitting that out and mimic the DynamoDBTableItem like you suggested.

sas-pemcne:
I moved the changes to the new resource of CognitoUserPoolDomain, same functionality as before and tested locally.

Sample output of scan:
```
us-east-2 - CognitoUserPool - pemcne-test-1 - would remove
us-east-2 - CognitoUserPool - pemcne-test-2 - would remove
us-east-2 - CognitoUserPoolDomain - pemcne-test-2 -> pemcne-test-2-domain - would remove
Scan complete: 3 total, 3 nukeable, 0 filtered.
```
And sample output of remove which shows the initial error and eventual consistency:
```
us-east-2 - CognitoUserPool - pemcne-test-1 - triggered remove
us-east-2 - CognitoUserPool - pemcne-test-2 - InvalidParameterException: User pool cannot be deleted. It has a domain configured that should be deleted first.
	status code: 400, request id: 5b0d8f12-f3dc-11e8-affd-cfb405cd0948
us-east-2 - CognitoUserPoolDomain - pemcne-test-2 -> pemcne-test-2-domain - triggered remove

Removal requested: 2 waiting, 1 failed, 0 skipped, 0 finished

us-east-2 - CognitoUserPool - pemcne-test-1 - waiting
us-east-2 - CognitoUserPool - pemcne-test-2 - triggered remove
us-east-2 - CognitoUserPoolDomain - pemcne-test-2 -> pemcne-test-2-domain - waiting

Removal requested: 3 waiting, 0 failed, 0 skipped, 0 finished

us-east-2 - CognitoUserPool - pemcne-test-1 - removed
us-east-2 - CognitoUserPool - pemcne-test-2 - waiting
us-east-2 - CognitoUserPoolDomain - pemcne-test-2 -> pemcne-test-2-domain - removed

Removal requested: 1 waiting, 0 failed, 0 skipped, 2 finished

us-east-2 - CognitoUserPool - pemcne-test-2 - removed

Removal requested: 0 waiting, 0 failed, 0 skipped, 3 finished

Nuke complete: 0 failed, 0 skipped, 3 finished.
```

PR #308, opened 2018-12-04
Added IAMServiceSpecificCredentials resource

sas-pemcne:
I was hitting an error with the nuker involving IAM user dependencies:

```
global - IAMUser - test - DeleteConflict: Cannot delete entity, must remove referenced objects first.
	status code: 409
```
This is due to service specific credentials being attached to the user that need to be deleted first. I have added a new resource called IAMServiceSpecificCredential which gathers all of these keys from IAM users. 

I included the credential's service (gathered from the Describe api call) to include in the nuker output. I think codecommit is the only service right now but I thought being verbose in the output would be good in case they add more services in the future. Output will be the following in `iam-user -> [cred-service] service-credential-name` format.
```
global - IAMServiceSpecificCredential - test -> [codecommit.amazonaws.com] test-at-123456789012 - would remove
```

### DIFF
--- a/resources/iam-service-specific-credentials.go
+++ b/resources/iam-service-specific-credentials.go
@@ -1,0 +1,74 @@
+package resources
+
+import (
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/sirupsen/logrus"
+)
+
+type IAMServiceSpecificCredential struct {
+	svc         *iam.IAM
+	name        string
+	serviceName string
+	id          string
+	userName    string
+}
+
+func init() {
+	register("IAMServiceSpecificCredential", ListServiceSpecificCredentials)
+}
+
+func ListServiceSpecificCredentials(sess *session.Session) ([]Resource, error) {
+	svc := iam.New(sess)
+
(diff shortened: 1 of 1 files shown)


resources/iam-service-specific-credentials.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+		params := &iam.ListServiceSpecificCredentialsInput{
+			UserName: &user.name,
+		}
+		serviceCredentials, err := svc.ListServiceSpecificCredentials(params)
+		if err != nil {
+			return nil, err
+		}
+
+		// ServiceSpecificCredentials comes back as an array
```
svenwltr: IMO this comment does not clarify very much.

sas-pemcne:
Good suggestion about the properties, I'll look to include that.

sas-pemcne:
I added the ID and the ServiceName to the properties. Since the properties are included in the output, I cleaned up the `String()` function to reflect that. 

Output will now look similar to:
```
global - IAMServiceSpecificCredential - test -> test-at-123456789012 - [ServiceName: "codecommit.amazonaws.com", ID: "ASDFASDFASDFADSF"] - would remove
```

svenwltr:
Thank you very much!

PR #310, opened 2018-12-11
Added resource EC2VPCEndpointServiceConfiguration

sas-pemcne:
I have added a new resource EC2VPCEndpointServiceConfiguration that will support private links within the VPC. These are different from the VPC Endpoints which there is already a resource. I tried to mimic the EC2VPCEndpoint resource as much as possible but just wanted to note that filtering by `vpc-id` is not supported in the [DescribeVpcEndpointServiceConfigurations API call](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeVpcEndpointServiceConfigurations.html). This does mean that VPC filters on the nuke configuration will not exclude this resource, let me know if that's an issue and I can try and figure out a work around. 

The resource name is quite long but I wanted to stay consistent with the project and the APIs. Let me know if you have any alternative suggestions.

This also fixes the issue where NLBs are tied to a private link and unable to be deleted:

```
us-east-1 - ELBv2 - foo-nlb - [] - ResourceInUse: Load balancer 'arn:aws:elasticloadbalancing:us-east-1:123456789012:loadbalancer/net/foo-nlb/123456' cannot be deleted because it is currently associated with another service
	status code: 400
us-east-1 - ELBv2TargetGroup - foo-tg - [] - ResourceInUse: Target group 'arn:aws:elasticloadbalancing:us-east-1:123456789012:targetgroup/foo-tg/123456' is currently in use by a listener or a rule
	status code: 400
```

I tied the name property in with the resource as well so the output will look similar to the following: 

```
us-east-2 - EC2VPCEndpointServiceConfiguration - vpce-svc-01f2ce46fabd7 - [Name: "com.amazonaws.vpce.us-east-2.vpce-svc-01f2ce46fabd7"] - would remove
```

### DIFF
--- a/resources/ec2-vpc-endpoint-service-configurations.go
+++ b/resources/ec2-vpc-endpoint-service-configurations.go
@@ -1,6 +1,7 @@
 package resources
 
 import (
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/rebuy-de/aws-nuke/pkg/types"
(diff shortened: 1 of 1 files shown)


svenwltr:
> These are different from the VPC Endpoints which there is already a resource.

This is perfectly fine. If AWS has a separate endpoint, we should have a separate resource for this.

> that filtering by vpc-id is not supported in the DescribeVpcEndpointServiceConfigurations API call

As fas as I see the whole resource is not bound to a VPC at all, right? This means filtering by VPC would just make no sense.

> The resource name is quite long 

They are based on the AWS names, so this is perfectly fine.

Issue #302, opened 2018-11-23
Support peering connections


n3ph:
Hi there!

I would like to nuke peering connections...
Is there something already in the pipeline? Otherwise I will have a look for over the weekend...

Best regards,

n3ph

svenwltr:
Hi @n3ph, we did not plan to do this yet. A PR would be welcome.

PR #314, opened 2019-01-22
VPC Peering Connections

MrPrimate:
Add Support for deleting VPC Peering Connections as per #302

### DIFF
--- a/resources/ec2-vpc-peering-connections.go
+++ b/resources/ec2-vpc-peering-connections.go
@@ -1,0 +1,61 @@
+package resources
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+)
+
+type EC2VPCPeeringConnection struct {
+	svc *ec2.EC2
+	id  *string
+}
+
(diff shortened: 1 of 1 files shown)


resources/ec2-vpc-peering-connections.go
```diff
@@ -0,0 +1,61 @@
...
+	register("EC2VPCPeeringConnection", ListEC2VPCPeeringConnections)
+}
+
+func ListEC2VPCPeeringConnections(sess *session.Session) ([]Resource, error) {
+	svc := ec2.New(sess)
+	resources := make([]Resource, 0)
+
+	// filter should be set as deleted vpc connetions are returned
+	params := &ec2.DescribeVpcPeeringConnectionsInput{
+		Filters: []*ec2.Filter{
```
svenwltr: For consistency, the filters should be applied in the `Filter` function, like done [with EC2 instances](https://github.com/rebuy-de/aws-nuke/blob/2215621/resources/ec2-instances.go#L42-L47).

As far as I remember we did it this way, so we can still see all resources in the output, regardless of it being filtered or not.

MrPrimate: Thanks for the feedback, I have amended accordingly.

Issue #188, opened 2018-05-03
Ability to exclude Default VPC


leozhad:
I would love to be able to exclude the default VPC via the config file

svenwltr:
Will be implemented in #198. The filter should look like this:

```yaml
accounts:
  "1234567890":
    filters:
      EC2VPC:
      - property: IsDefault
        value: "true"
```

svenwltr:
This should be fixed in [v2.1.0-beta.1](https://github.com/rebuy-de/aws-nuke/releases/tag/v2.1.0-beta.1).

PR #315, opened 2019-01-22
Add 'DefaultForAz' property for subnets

TomAnthony:
Similar to issue https://github.com/rebuy-de/aws-nuke/issues/188 (implemented in https://github.com/rebuy-de/aws-nuke/pull/198), which allowed for easy filtering of the default VPC, this does the same for default subnets.

You can use it like so:

```yaml
accounts:
  "1234567890":
    filters:
      EC2Subnet:
      - property: DefaultForAz
        value: "true"
```


svenwltr:
Thanks!

PR #319, opened 2019-01-28
Remove All attached MFA devices and SSH Public keys before removing an IAMUser

jakubincloud:
Updated IAMUser resource to:
detach all mfa devices attached to a given user
delete SSH public keys created to a given user
This is required to fully remove the IAM User, otherwise we get an error:
"DeleteConflict: Cannot delete entity, must remove referenced objects first. status code: 409"

### DIFF
--- a/resources/iam-users.go
+++ b/resources/iam-users.go
@@ -1,6 +1,7 @@
 package resources
 
 import (
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/iam"
 )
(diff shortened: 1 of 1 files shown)


svenwltr (commented):
Not sure if MFADevices and SSHPublicKeys should be their own resources. Thoughts @rebuy-de/prp-aws-nuke?

dxciberaws:
+1 
For consistency with AWS reference documentation, MFA is an IAM resource type, but SSH keys in IAM are a user's property.

https://docs.aws.amazon.com/IAM/latest/UserGuide/list_identityandaccessmanagement.html

svenwltr:
> Not sure if MFADevices and SSHPublicKeys should be their own resources. Thoughts @rebuy-de/prp-aws-nuke?

This is still unaddressed. Due to the age I am closing this.

PR #326, opened 2019-02-06
add --force-sleep param instead of statically setting the wait to 15 seconds

justinclayton:
this is most useful if run as part of a lambda, where sleeps == $$.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -105,7 +105,11 @@ func NewRootCommand() *cobra.Command {
 	command.PersistentFlags().BoolVar(
 		&params.Force, "force", false,
 		"Don't ask for confirmation before deleting resources. "+
-			"Instead it waits 15s before continuing.")
+			"Instead it waits 15s before continuing. Set --force-sleep to change the wait time.")
+	command.PersistentFlags().IntVar(
+		&params.ForceSleep, "force-sleep", 15,
+		"If specified and --force is set, wait this many seconds before deleting resources. "+
+			"Defaults to 15.")
 	command.PersistentFlags().IntVar(
 		&params.MaxWaitRetries, "max-wait-retries", 0,
 		"If specified, the program will exit if resources are stuck in waiting for this many iterations. "+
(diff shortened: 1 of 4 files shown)


cmd/root.go
```diff
@@ -105,7 +105,11 @@ func NewRootCommand() *cobra.Command {
 	command.PersistentFlags().BoolVar(
 		&params.Force, "force", false,
 		"Don't ask for confirmation before deleting resources. "+
-			"Instead it waits 15s before continuing.")
+			"Instead it waits 15s before continuing. Set --force-sleep to change the wait time.")
+	command.PersistentFlags().IntVar(
```
svenwltr: There is a function that outputs a duration directly: https://godoc.org/github.com/spf13/pflag#DurationVar

bjoernhaeuser:
@svenwltr What do you think about allowing to set, but keep at least at 1 second?

svenwltr:
> What do you think about allowing to set, but keep at least at 1 second?

Still feels a bit short to me.

bjoernhaeuser:
It is short, but otherwise if aws-nuke is running inside lambda it makes sense to me.

svenwltr:
Ok, since there are no other opinions, I will accept a minimum of 1 second.

justinclayton:
@svenwltr Just for some context, running this in two regions on an account that was already cleaned seems to take about 15-20 seconds for me, not counting the sleep time. Since my use case is sandbox accounts used for training, I am fine running in a non-interactive mode.

That said, I'm happy to create a minimum for the "wait no no no no" use case. I did some non-scientific simulations of this use case and found that I agree with you 1 second is indeed too fast for the expected human interaction (i.e., frantically finding the ctrl and c keys). 3 seconds felt like a much better minimum to me. Unless there are any objections, I'll add a commit that adds that minimum.

cmd/root.go
justinclayton: True, but doing this forces the user to know to type "15s" instead of "15". If we foresee users wanting to set durations that aren't easily convertible into seconds (60, 300, etc.), then it makes sense, otherwise I feel it makes more sense to just use seconds.

bjoernhaeuser:
@svenwltr merge?

PR #335, opened 2019-02-22
ensure ELB deletion works with > 20 load balancers; remove all S3 object versions

steved:
* `LoadBalancerNames` allows only a maximum of 20 names in a list before bombing out, so call the `DescribeTags` function on each load balancer in a loop even though it's more inefficient.

* Removing all S3 objects is unfortunately not enough if versioning is enabled. This PR empties all object versions as well.


resources/s3-object-versions.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"github.com/aws/aws-sdk-go/aws/request"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3iface"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+)
+
+// DeleteListVersionListIterator is an iterator for S3 Object Versions.
+type DeleteVersionListIterator struct {
```
svenwltr: The type name should contain `S3`, so the name makes clear what service the resource belongs to. Also the file name should match the type name.

Since the type is only used by `S3Bucket` I think it should even be part of `s3-buckets.go`, since we should keep the number of files low.

steved: Makes sense. I actually originally had it in `s3-buckets.go`, I'll move it back.

PR #340, opened 2019-03-01
filter inline role policies for roles created by AWS

### LINKED ISSUE #331: AWSServiceRoleForAmazonGuardDuty should be excluded
Using AWS Nuke v2.7.0, we encountered this error:

`global - IAMRolePolicy - AWSServiceRoleForAmazonGuardDuty -> b4b3b0b27e29f5db632f1c5405f658e3 - UnmodifiableEntity: Cannot perform the operation on the protected role 'AWSServiceRoleForAmazonGuardDuty' - this role is only modifiable by AWS
`

svenwltr:
> Fixes #331

Additionally, it adds all properties to IAMRolePolicy resources.

### DIFF
--- a/resources/iam-role-policy.go
+++ b/resources/iam-role-policy.go
@@ -2,14 +2,17 @@ package resources
 
 import (
 	"fmt"
+	"strings"
 
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/rebuy-de/aws-nuke/pkg/types"
 )
 
 type IAMRolePolicy struct {
 	svc        *iam.IAM
-	roleName   string
+	role       iam.Role
 	policyName string
 }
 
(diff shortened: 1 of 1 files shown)


resources/iam-role-policy.go
```diff
@@ -65,10 +68,17 @@ func ListIAMRolePolicies(sess *session.Session) ([]Resource, error) {
 	return resources, nil
 }
 
+func (e *IAMRolePolicy) Filter() error {
+	if strings.HasPrefix(aws.StringValue(e.role.Path), "/aws-service-role/") {
+		return fmt.Errorf("cannot alter service roles")
```
bjoernhaeuser: why is this an error?

svenwltr: Back then I thought it would be smart to design the interface like this ...

bjoernhaeuser: Understood. Though we could log here and still return `nil`.

svenwltr: That would not work, because we want the control loop to know that it should not attempt to delete this resource.

PR #349, opened 2019-04-11
fix DynamoDBItem listing

svenwltr:
Fixes this error:

```
Listing DynamoDBTableItem failed:
    ValidationException: Invalid ProjectionExpression: Attribute name is a reserved keyword; reserved keyword: Path
```


svenwltr:
> I wonder if there are more reserved keywords...

I wonder why we are using projections, if it apparently also works without ¯\_(ツ)_/¯

PR #352, opened 2019-04-17
Let aws-nuke container run as non-root user

der-eismann:


svenwltr:
Not sure how much of a breaking change this is. If someone copies a config into the container, which is not readable by the new user, it will break.

der-eismann:
That is true. I would argue that normally all files are at least `644`, but I get that aws credentials are a bit sensitive. So... what do?

svenwltr:
I guess adding a `Breaking Change` section to the next release should be enough.

Issue #355, opened 2019-04-23
MaxRetriesExceed does not return none zero exit code


videte47:
This is sort of a question, but I would think that when resources exceed the wait retries threshold aws-nuke should exit with a none zero code no ? This makes the caller aware that there are some orphan resources that are left behind. Currently, aws-nuke while printing the message that we have exceeded the threshold it exits cleaning making the caller assumed that the job was properly completed and all resources were removed. 
https://github.com/rebuy-de/aws-nuke/blob/master/cmd/nuke.go#L117-L118
I wanted to put this issue to get some feedback @svenwltr before submitting a PR

svenwltr:
Yes, you are right @videte47. It should return an error instead of just exiting the loop here.

videte47:
Closing issue per https://github.com/rebuy-de/aws-nuke/pull/356

PR #356, opened 2019-04-24
return error when max wait threshold is reached

videte47:
This is in response to #355. aws-nuke now returns error when threshold is reached

### DIFF
--- a/cmd/nuke.go
+++ b/cmd/nuke.go
@@ -115,9 +115,7 @@ func (n *Nuke) Run() error {
 		}
 		if n.Parameters.MaxWaitRetries != 0 && n.items.Count(ItemStateWaiting, ItemStatePending) > 0 && n.items.Count(ItemStateNew) == 0 {
 			if waitingCount >= n.Parameters.MaxWaitRetries {
-				fmt.Printf("Max wait retries of %d exceeded.\n\n", n.Parameters.MaxWaitRetries)
-
-				break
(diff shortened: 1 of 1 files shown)


cmd/nuke.go
```diff
@@ -115,9 +115,7 @@ func (n *Nuke) Run() error {
 		}
 		if n.Parameters.MaxWaitRetries != 0 && n.items.Count(ItemStateWaiting, ItemStatePending) > 0 && n.items.Count(ItemStateNew) == 0 {
 			if waitingCount >= n.Parameters.MaxWaitRetries {
-				fmt.Printf("Max wait retries of %d exceeded.\n\n", n.Parameters.MaxWaitRetries)
-
-				break
+				return fmt.Errorf("Max wait retries of %d exceeded.\n\n", n.Parameters.MaxWaitRetries)
```
bjoernhaeuser: I am not 100% about this. This would mean the first time this condition is reached the loop is terminated.

bjoernhaeuser: I think the handling should happen in line 134, right?

svenwltr: > This would mean the first time this condition is reached the loop is terminated.

This is the idea. If there are more than `MaxWaitRetries` iterations that contain resources in pending state, it prematurely exits.

> I think the handling should happen in line 134, right?

I would say both ways would be okay, but it should be consistent with Line 109.

videte47:
Hey guys, @bjoernhaeuser @svenwltr what do you guys recommend that I amend to this - thanks for the feedback

PR #364, opened 2019-05-10
Initial version for snapcraft

der-eismann:
Seems like this creates a working snap, but I'm not sure how to automate builds and releases.
@svenwltr

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,0 +1,19 @@
+name: aws-nuke
+version: 2.10.0
+summary: Nuke a whole AWS account and delete all its resources.
+description: |
+  Nuke a whole AWS account and delete all its resources.
+license: MIT
+
+confinement: classic
+grade: devel
+base: core18
+
+parts:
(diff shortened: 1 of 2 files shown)


snapcraft.yaml
```diff
@@ -0,0 +1,19 @@
...
+base: core18
+
+parts:
+  aws-nuke:
+    plugin: dump
+    source: https://github.com/rebuy-de/aws-nuke/releases/download/v2.10.0/aws-nuke-v2.10.0-linux-amd64.tar.gz
+
+apps:
+  aws-nuke:
+    command: aws-nuke-v2.10.0-linux-amd64
```
stephanlindauer: so the next version would have to be appended in this file?

der-eismann: No, with every release we would need to:
- edit the version at the top
- change the tarball link
- change the command (which is the unpacked binary)

Would be cool if we could automate that, but I'm not sure how.

PR #365, opened 2019-05-12
Enable tags for EC2 DHCP Options

hmalphettes:
Many thanks for a very useful project.

I ported https://github.com/rebuy-de/aws-nuke/commit/4dfad10643e291fa3353b82fe23ffe614f0b91f0 for EC2 DHCP Options.

I am actually using this patch.


videte47:
Do you think is worth adding an example under the config/example.yaml for ease of reference in the future?

svenwltr:
> Do you think is worth adding an example under the config/example.yaml for ease of reference in the future?

I would say the example file should just show the idea. If we add every resource there, the file will get very big and unclear.

hmalphettes:
Cheers to both of you!

PR #371, opened 2019-05-19
Support for custom endpoints

hmalphettes:
This patch provides the ability to use aws-nuke against non default AWS endpoint.
Terraform's aws provider supports custom endpoints here: https://www.terraform.io/docs/providers/aws/guides/custom-service-endpoints.html

It could be used for integration testing pointing at "local stack".
In my case, I am pointing aws-nuke at Stratoscale clusters.

I am not very happy with the current implementation in this PR, although it does work for our use case.
- defining new AWS partition does not really seem to be possible (let me know if I missed something)
- terraform's implementation is directly at the session level for each service type: https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/config.go
- the patch will either make a session for each service type of a region where custom endpoints are defined or it will fall back to use the standard mechanism.
- for my usecase, there is no support for account aliasing, so I do detect that and present a prompt where the identifier of the region needs to be entered as if it was on of the aliases

If it is useful for the community let me know. Happy to refactor according to your guidance and in any case, thanks again for very useful tool.


svenwltr:
Hello @hmalphettes. Thanks for your contribution. Is there an easy way to test this? I have concerns that we might break this feature in future commits, if we simply cannot verify it.

hmalphettes:
Hi @svenwltr - I am not really sure actually.
Terraform actually documents:

> NOTE: Support for connecting the Terraform AWS Provider with custom endpoints and AWS compatible solutions is offered as best effort. Individual Terraform resources may require compatibility updates to work in certain environments. Integration testing by HashiCorp during provider changes is exclusively done against default AWS endpoints at this time.

I actually don't have an AWS account active at the moment.

We could definitely test scanning an AWS account using the default mechanism and using the custom endpoints: as long as we get the same list of resources back, I think we are good to go.

If this makes sense, I'll add an example configuration for that.

svenwltr:
@hmalphettes This would be fine for me, if we add a note to the README, that endpoints are only supported as best effort and having an this example.

hmalphettes:
Hi @svenwltr and sorry for dropping the ball for so long.
I was content enough for my own use.
The engineers of Stratoscale are using the custom endpoints found on this PR.
Would you still consider merging this and let -me- us know what would be missing?
Thanks a lot!

rom-stratoscale:
@hmalphettes lets rename demo10 to stratoscale :)

svenwltr:

README.md
```diff
@@ -207,6 +207,87 @@ or in [shared config
 file](https://docs.aws.amazon.com/cli/latest/userguide/cli-roles.html) with an
 assuming role.
 
+### Using custom AWS endpoint
+
+It is possible to configure aws-nuke to against non default AWS endpoints.
```
der-eismann:
```suggestion
It is possible to configure aws-nuke to run against non-default AWS endpoints.
```

README.md
```diff
@@ -207,6 +207,87 @@ or in [shared config
...
+It could be used for integration testing pointing to a local endpoint such as an
+S3 appliance or a Stratoscale cluster for example.
+
+To configure aws-nuke to use custom endpoint, add the configuration directives as shown in the following example:
```
der-eismann:
```suggestion
To configure aws-nuke to use custom endpoints, add the configuration directives as shown in the following example:
```

cmd/root.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"sort"
```
der-eismann:
```suggestion
```

pkg/awsutil/session.go
```diff
@@ -1,21 +1,29 @@
 package awsutil
 
 import (
+	"crypto/tls"
 	"fmt"
 	"net"
 	"strings"
```
der-eismann:
```suggestion
```

bjoernhaeuser:
@hmalphettes thank you very much for the contribution, will be in the next release. <3

PR #372, opened 2019-05-22
Added tags list for vpn connections

a-teisseire:
Hello,

This quick and easy PR's sole purpose is to add tags to the VPN connection object.
Let me know if you have any remark.

Kind regards,

### DIFF
--- a/resources/ec2-vpn-connections.go
+++ b/resources/ec2-vpn-connections.go
@@ -5,12 +5,12 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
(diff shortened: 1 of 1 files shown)


resources/ec2-vpn-connections.go
```diff
@@ -2,15 +2,15 @@ package resources
 
 import (
 	"fmt"
+	"github.com/rebuy-de/aws-nuke/pkg/types"
```
der-eismann: Please but this in a block with the other GitHub packages

a-teisseire: Should be okay now. I have rebased my commit to avoid having another one just for this fix.

Issue #377, opened 2019-06-13
Can't get docker running


schmidalex37:
So i can't get the docker running. The config works fine and the programm can be executed locally with go. 
`go run main.go --config config/nuke-settings.yaml --profile testUser --force --no-dry-run`

Configurations are applied correctly.
`docker build . -t rebuy-de/aws_nuke/playground-nuke` 
works fine and the build is successfully built

The problem is if i try to run the docker this way i get this error
```
docker run -a stdin -a stdout -ti rebuy-de/aws_nuke/playground-nuke -c config/nuke-config.yml --profile testUser -v
DEBU[0000] creating new session in us-east-1
DEBU[0000] creating new root session in us-east-1
Error: failed get caller identity: NoCredentialProviders: no valid providers in chain. Deprecated.
        For verbose messaging see aws.Config.CredentialsChainVerboseErrors
``` 
but i can get the caller identity with this code:
```
aws sts get-caller-identity --profile testUser
{
    "Account": "############################",
    "UserId": "############################",
    "Arn": "############################"
}
```
if i use the key directly in the console this happens, even if i use the example config file it does not work:

```
C:\Go\src\github.com\rebuy-de\aws-nuke\playground-nuke>docker run -a stdin -a stdout -ti rebuy-de/aws_nuke/playground-nuke -c config/nuke-settings.yaml --access-key-id=########  --secret-access-key=######## --session-token=######## -v
DEBU[0000] creating new session in us-east-1
DEBU[0000] creating new root session in us-east-1
DEBU[0000] sending AWS request:
    > POST / HTTP/1.1
    > Host: sts.amazonaws.com
    > Authorization: <hidden>
    > Content-Length: 43
    > Content-Type: application/x-www-form-urlencoded; charset=utf-8
    > User-Agent: aws-sdk-go/1.19.48 (go1.12.5; linux; amd64)
    > X-Amz-Date: 20190613T071439Z
    > X-Amz-Security-Token: <hidden>
    >
    > Action=GetCallerIdentity&Version=2011-06-15
DEBU[0000] received AWS response:
    < HTTP/1.1 200 OK
    < Content-Length: 469
    < Content-Type: text/xml
    < Date: Thu, 13 Jun 2019 07:14:39 GMT
    < X-Amzn-Requestid: e89d725d-8daa-11e9-930d-77a502702f1e
    <
    < <GetCallerIdentityResponse xmlns="https://sts.amazonaws.com/doc/2011-06-15/">
[...]

PR #379, opened 2019-06-17
clearify Docker section in the readme

svenwltr:
> Follow-up to https://github.com/rebuy-de/aws-nuke/issues/377

### DIFF
--- a/README.md
+++ b/README.md
@@ -460,22 +460,28 @@ directory or `make install` go install *aws-nuke* into `$GOPATH/bin`. With
 
 ### Docker
 
-To use the provided Docker implementation, first you need to make sure you've 
-created the config file and that it exists in the `config` directory.
-Make sure you are in the project's root directory, then you need to build 
-the image and run the container:
+You can run *aws-nuke* with Docker by using a command like this:
 
 ```bash
-$ docker build . -t rebuy-de/aws_nuke
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -460,22 +460,28 @@ directory or `make install` go install *aws-nuke* into `$GOPATH/bin`. With
...
 
 ```bash
-$ docker build . -t rebuy-de/aws_nuke
-
-$ docker run -a stdin -a stdout -ti rebuy-de/aws_nuke -c config/nuke-config.yml --profile aws-nuke-example
+$ docker run \
+    --rm -it \
+    -v /full-path/to/nuke-config.yml:/home/aws-nuke/config.yml \
+    -v /home/user/.aws:/home/aws-nuke/.aws \
+    quay.io/rebuy/aws-nuke:v2.11.0 \
```
bjoernhaeuser:
```suggestion
    quay.io/rebuy/aws-nuke:latest \
```

Otherwise you will always have to update the README

svenwltr: We only have `master` and I somehow would not like to suggest using it.

We would need to update the CI, if we want to support `latest`.

der-eismann: A latest versions would be nice, not sure how hard it is to implement. We can't expect users to check the `Releases`  or quay.io page and change it accordingly.

PR #380, opened 2019-06-24
Delete secrets without recovery window instead of scheduling a deletion

kurtmc:


kurtmc:
I ran into an issue with this. After running aws-nuke I tried to deploy a terraform stack and I could not because there where secrets with the same name but scheduled for deletion. This should resolve that issue.

PR #384, opened 2019-07-11
Format resource properties in sorted order

paulgear:
This ensures that multiple runs of aws-nuke produce stable output so
that they can be compared with each other.

This is likely terrible golang style/design.  Happy to take advice on how to make it more idiomatic.

### DIFF
--- a/cmd/log.go
+++ b/cmd/log.go
@@ -2,6 +2,8 @@ package cmd
 
 import (
 	"fmt"
+	"sort"
+	"strings"
 
 	"github.com/fatih/color"
 	"github.com/rebuy-de/aws-nuke/resources"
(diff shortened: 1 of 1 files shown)


paulgear:
> For future commits it would be cool if your commits actually match your GitHub user (I think the mail address is missing) slightly_smiling_face

Apologies for that - I was committing from work where I don't have my personal ssh key, but used an API token instead.  Do you prefer GPG-signed commits?

paulgear:
Any further action this needs from me?

der-eismann:
It's not about SSH or GPG keys, it's just that your email of your commit (github@libertysys.com.au) probably isn't your GitHub email address. On the [page of your commit](https://github.com/rebuy-de/aws-nuke/pull/384/commits/1cafa255af81e11ec7348b01862ce13330493b13) you can see that your avatar is missing and doesn't link to your profile. 
If your git config at work differs from your private one you should have a look at [this](https://smarterco.de/set-the-username-and-email-in-git-globally-and-per-project/).

After changing this we will merge it.

svenwltr:
Sorry for the late reply. As @der-eismann said, it is just about associating the commit authors email with GitHub, but I think this is not a blocker.

PR #388, opened 2019-07-17
Fix: catch s3 bucket tag error properly

sambattalio:
Untagged buckets get returned an error from GetBucketTagging API call. So, buckets without tags currently do not get displayed. This fix now will only skip buckets on actual api call errors.

Sorry I didn't catch this when I pushed these changes earlier :/


sambattalio:
Thanks for the help, I think this looks a lot better haha. What do you think?

Issue #374, opened 2019-05-27
Fail to remove IoT Resources


EliassonHenrik:
I've noticed that the following targets have issues, preventing IoT-resources to be deleted:
- IoTCertificate
- IoTPolicy
- IoTThing (x2)

In other words, all of these targets needs to implement some preparations before the actual delete can be performed.

**IoTCertificate**
Certificates can not be deleted if they are `ACTIVE`, they need to be set to `INACTIVE` before.
If not, `DeleteCertificate` action get this error:
```
CertificateStateException
Certificate must be deactivated (not ACTIVE) before deletion. Id: <certificate>
```

**IoTPolicy**
Policies can not be deleted if they have certificates attached, they need to be detached before.
If not, `DeletePolicy` action get this error:
```
DeleteConflictException
The policy cannot be deleted as the policy is attached to one or more principals (name=<policy>)
```

**IoTThing**
Things can not be deleted if they have certificates attached, they need to be detached before. 
If not, `DeleteThing` action get this error:
```
InvalidRequestException
Cannot delete. Thing <thing> is still attached to one or more principals
```

**IoTThing**
Things can not be deleted if they have versions available, versions need to be deleted before. 
If not, `DeletePolicy` action get this error:
```
DeleteConflictException
The policy cannot be deleted as versions other than than default version exist (name=<policy>)
```

bjoernhaeuser:
@EliassonHenrik would you be interested in providing the necessary changes?

npellegrin:
Was the issue fixed by #392 ? If yes, this issue could be closed.

der-eismann:
Fixed by #392

PR #392, opened 2019-07-29
Adds additional AWS IoT considerations for deletion

richardneililagan:
This PR addresses the points detailed in #374. 

Changes:
- Now detaches attached principals to IoT things prior to deletion
- Now detaches attached targets to IoT policies prior to deletion
- Now deprecates non-default versions of IoT policies prior to deletion
- Now deactivates IoT certificates prior to deletion

### DIFF
--- a/resources/iot-certificates.go
+++ b/resources/iot-certificates.go
@@ -7,9 +7,8 @@ import (
 )
 
 type IoTCertificate struct {
-	svc    *iot.IoT
-	ID     *string
-	status *string
+	svc *iot.IoT
+	ID  *string
 }
 
 func init() {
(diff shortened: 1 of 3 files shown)


timboven:
Just tested this one out as I had the same issue and this seemed to work fine

svenwltr:
Hello and sorry to the delay.

I think it would be more clean to exclude those resources within the filter function. See [EC2 Instance](https://github.com/rebuy-de/aws-nuke/blob/08e6b8313b510f7d43fe5b45a090ed48c98bfc6a/resources/ec2-instances.go#L42-L47) for an example.

The reason for doing so is to give the user a proper expectation what resources exist and what *aws-nuke* is doing with them in the initial resource listing.

richardneililagan:
@svenwltr Apologies --- have been away for some R&R.

I'd appreciate it if you could expound on that more --- there's essentially no hard filtering involved in the PR; really more of just additional processing for some resources, but in the end, the resources still get nuked.

svenwltr:
@richardneililagan You are right. Thanks for your patience.

PR #417, opened 2019-10-01
Terminate protected EC2 instances

jbmchuck:
Currently EC2 instances which have `DisableApiTermination` set
`true` are not terminated.

A few other types (rds instances, clusters) handle this case by
disabling protection if it is detected in the `List..` function.
Unfortunately in the case of EC2 instances the protection flag
(`DisableApiTermination`) is not returned by the `DescribeInstances`
call.

This leaves a few options to force termination of instances which have
`DisableApiTermination` set `true`. Assuming `n` is the total number
of EC2 instances and `m` is the number of instances with protection
enabled and `m <= n` will always hold.

* Call `DescribeInstanceAttribute` per instance inside
  `ListEC2Instances` and `ModifyInstanceAttribute` for each account
  where protection is enabled. This approach results in `n+m`
  additional API calls regardless of status.
* Always call `ModifyInstanceAttribute` in `ListEC2Instances`. This
  results in `n` additional API calls regardless of status.
* Call `ModifyInstanceAttribute` only if `TerminateInstances` fails
  with an error indicating instance protection prevented termination.
  This results in `2m` additional api calls - for each protected instance a modify and an additional terminate. 

This PR takes the 3rd approach which minimizes API calls if 1/2 or less of total instances are protected.  Best case it adds no additional calls, worst case it adds 2n calls.  This seems like a reasonable solution to me for most environments.


svenwltr:
@rebuy-de/prp-aws-nuke Do you have any thoughts on this?

bjoernhaeuser:
I like very much! (Also sorry for the late answer). Though I would still put that feature behind an opt-in.

jbmchuck:
Thanks - is there currently an existing mechanism in system to do a per-resource-type flag like this?

jbmchuck:
Perfect, thanks - looks great.

svenwltr:
@jbmchuck It would be nice, if you use those feature flags. If you are short on time, I could do this for you.

jbmchuck:
PR updated to use the new feature flags feature.

Ran a few tests in my environment, looked good.

svenwltr:
Thank you!

PR #424, opened 2019-10-30
add tests and improve support for cloudformation stack delete

tylersouthwick:
Fix cloudformation stack deletion to wait for stack to complete before finishing. This adds unit tests around cloudformation stack delete, bringing in gomock to generate mocks.


svenwltr:
Hello @tylersouthwick. Thanks for the PR and sorry for the late response. It took a while, because I had to think about the tests. The thing is, that the tests only verify, that the code does not change, but it does not guarantee that the resource will still work in the future. When AWS changes the API or the behavior a little bit, things might break anyways.

I think the only way to really test this is to do the tests against the API. We started to work on this [a long time ago](https://github.com/rebuy-de/aws-nuke/pull/177), but never finished it.

Additionally, using `gomock` raises the entry bar for new developers, because they need to install yet another tool and need to make sure it is a compatible version.

Any thoughts on this @rebuy-de/prp-aws-nuke and @tylersouthwick?

bjoernhaeuser:
I would vote for removing gomock. We anyways have to test it against the "real" AWS APIs.

tylersouthwick:
While I agree there is value in testing against the real APIs, there is also value in unit tests. Testing only against the real apis are expensive and slow, let alone scalable. At some point, the api contract has to be trusted. One of my big concerns in this project is the lack of unit tests; especially around non simple logic. Gomock is a standard way in go to generate mocks to verify behavior. It does require one to do a ‘go get’ which is annoying; I wish go mod could manage that. I am not sure why that would be a reason to not use a common tool.

bjoernhaeuser:
@tylersouthwick 

Sorry for the late response. I am trying to answer all your questions, also the ones you sent to Sven.

Are we going to reInvent?

No :)

Would this PR be already merged without mockgen/gomock?

Probably yes. We write a lot of go code these days and removed mockgen over two years from other (private) projects, because it always felt wrong in go and often made it more complex. 
That being said it still could make sense to have this here. 

Mockgen in general

I just tried to checkout your PR and “build” or “test” it locally. It does not work because for building the mocks are missing (why?) and are not automatically built (required changes in the Makefile missing) and for tests mockgen is missing.

So to further consider having mockgen here I think the minimum requirement would be:

make mockgen part of the build process
install it locally, if possible, when not yet installed
adjust readme

So the question is really do we want to maintain it?

Not really, though we are not heavy users of cloudformation (mostly terraform), so we will rely on community feedback (yours!) to adjust and maintain the code accordingly. So as long as mockgen will “stay” in the cloudformation part, we are fine, because you already made the neccessary changes and the tests are looking okay to us. 

Although we still have the opinion that mockgen adds a lot of complexity to this project and we will decide case by case (aka resource by resource) if we will accept further introduction.

tylersouthwick:
What command were you running where you couldn't build or test it?

bjoernhaeuser:
Running `make test` returns:

```
rm mocks -rvf
mockgen -source /Users/bjoernhaeuser/go/pkg/mod/github.com/aws/aws-sdk-go@v1.25.0/service/cloudformation/cloudformationiface/interface.go -destination mocks/mock_cloudformationiface/mock.go
make: mockgen: No such file or directory
make: *** [generate_mocks] Error 1
```

Running make:

```
go mod vendor
go: finding github.com/rebuy-de/aws-nuke/mocks latest
go: finding github.com/rebuy-de/aws-nuke/mocks/mock_cloudformationiface latest
github.com/rebuy-de/aws-nuke/resources tested by
	github.com/rebuy-de/aws-nuke/resources.test imports
	github.com/rebuy-de/aws-nuke/mocks/mock_cloudformationiface: no matching versions for query "latest"
make: *** [vendor] Error 1
```

tylersouthwick:
The readme is updated to include the need to do a ‘go get’. Could look into the makeflle automatically doing that. Thoughts?

tylersouthwick:
I found a way to use go generate that uses the go.mod dependencies so make test should work with no other configurations

Issue #435, opened 2019-11-25
v2.13.0 cannot get credentials when running on Lambda


mseiwald:
Hi,

we are running aws-nuke from a Python Lambda handler like this:

```
os.system("echo <account_name> | ./aws-nuke --profile default -c config.yaml --no-dry-run --force")
```

This has been working fine up until version `v2.11.0`. After upgrading to `v2.13.0` the execution fails to get AWS credentials.

```
Error: failed get caller identity: NoCredentialProviders: no valid providers in chain. Deprecated.
For verbose messaging see aws.Config.CredentialsChainVerboseErrors
```

Downgrading immediately resolved the issue. Is there some workaround to make it work on Lambda?

Regards
Michael

svenwltr:
Hello @mseiwald,

can you show us how the `default` profile looks like? Of course with redacted secrets.

mseiwald:
We are running on Lambda with an IAM role, there is no profile configured..

svenwltr:
Then, why are you specifying the `default` profile? Can you try this:

```
os.system("echo <account_name> | ./aws-nuke -c config.yaml --no-dry-run --force")
```

mseiwald:
When running without `--profile` we get this error:

```Error: You have to specify the --profile flag OR --access-key-id with --secret-access-key and optionally --session-token```

**Comment by svenwltr** [comment]
Ah, I see. We need to implement that *aws-nuke* falls back to EC2 Instance Roles, if there is no profile given.

**Comment by mseiwald** [comment]
Still strange that it worked fine before v2.13.0 even with `--profile`

**Comment by mseiwald** [comment]
When setting `CredentialsChainVerboseErrors` to `true` I get this error output:

```Error: failed get caller identity: NoCredentialProviders: no valid providers in chain
caused by: EnvAccessKeyNotFound: failed to find credentials in the environment.
SharedCredsLoad: failed to load profile, default.
EC2RoleRequestError: no EC2 instance role found
caused by: RequestError: send request failed
caused by: Get http://169.254.169.254/latest/meta-data/iam/security-credentials/: dial tcp 169.254.169.254:80: connect: connection refused
```

As I said the Lambda function has a role associated with it and it works flawlessly with `v2.11.0` so I assume it has something to do with the updated `aws-sdk-go`

mseiwald:
And I can confirm that `v2.13.0` with (downgraded) `aws-sdk-go v1.17.8` works. 🎉 

[...]

PR #438, opened 2019-12-11
allow to run without speficying the profile

svenwltr:
Should solve https://github.com/rebuy-de/aws-nuke/issues/435, but needs further testing.


mseiwald:
I can confirm that this branch works both on Lambda and locally when having credentials in the `AWS_*` environment variables.

PR #439, opened 2019-12-13
Fetch all Route53ResourceRecordSet

deanrock:
`ListResourceRecordSets` only fetches the first 100 records on each call. With this pull requests it will fetch all the records.


deanrock:
Sorry for force pushes, I made typo in the comment twice. :/

PR #450, opened 2020-01-12
Robomaker

mvanholsteijn:
Provides support for removing:

- robots
- fleets
- robot applications
- simulation applications

and canceling:

- deployment jobs
- simulation jobs
there is no delete operations on the latter two.

### DIFF
--- a/resources/robomaker-deployment-jobs.go
+++ b/resources/robomaker-deployment-jobs.go
@@ -1,0 +1,72 @@
+package resources
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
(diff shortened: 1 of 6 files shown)


resources/robomaker-deployment-jobs.go
```diff
@@ -0,0 +1,72 @@
...
+
+	_, err := f.svc.CancelDeploymentJob(&robomaker.CancelDeploymentJobInput{
+		Job: f.arn,
+	})
+
+	return err
+}
+
+func (f *RoboMakerDeploymentJob) String() string {
+	return *f.arn
```
bjoernhaeuser: is the arn descriptive enough?

mvanholsteijn: I think it is. The arn contains the name of the job, together with the account and region in which it was created.

PR #456, opened 2020-01-24
handle pagination in backup-plan and backup-selection resources

matthagenbuch:
I've found that in practice AWS Backup service will sometimes paginate backup plan results despite the MaxResults setting.  This will also handle future situations with >100 plans in an account.

### DIFF
--- a/resources/backup-plans.go
+++ b/resources/backup-plans.go
@@ -21,27 +21,36 @@ func init() {
 
 func ListBackupPlans(sess *session.Session) ([]Resource, error) {
 	svc := backup.New(sess)
-	false_value := false
-	max_backups_len := int64(100)
+	falseValue := false
+	maxBackupsLen := int64(100)
 	params := &backup.ListBackupPlansInput{
-		IncludeDeleted: &false_value,
-		MaxResults:     &max_backups_len, // aws default limit on number of backup plans per account
+		IncludeDeleted: &falseValue,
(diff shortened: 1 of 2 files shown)


nayyara-samuel:
I've seen the pagination in the AWS backup API cause confusion in other contexts and specifically for `aws-nuke`, since not all plans get destroyed, it can cause conflicts with tools we use such as Terraform. 
Would love to get this fix in!

resources/backup-plans.go
```diff
@@ -21,19 +21,30 @@ func init() {
...
+		MaxResults:     &maxBackupsLen, // aws default limit on number of backup plans per account
 	}
 	resp, err := svc.ListBackupPlans(params)
 	if err != nil {
 		return nil, err
 	}
 
+	backupPlansList := resp.BackupPlansList
+	nextToken := resp.NextToken
+	for nextToken != nil {
```
bjoernhaeuser: Thank you for your contribution!

In the other resources there is only one loop for fetching the resources, append to the resource slice and handling the next token (https://github.com/rebuy-de/aws-nuke/blob/master/resources/appstream-fleets.go#L17)

I cannot see a reason why not do the same thing here, therefore it would be great if you can change your code to look like the other resources.

Thank you!

matthagenbuch: Thanks for the suggestion, I've modified the change to match the example you provided.

matthagenbuch:
Note that I had to apply this to the backup-selection resource as well

PR #480, opened 2020-03-02
Feature/324 date filter

### LINKED ISSUE #324: Nuke resources based on Tags
Can we nuke resources based on Tags based on date.
For example, say my resource have tag of their expiry date like YYYY-MM-DD. So all resources that are running post this expiry date are to be nuked. Can we have any such tagging policy in Nuke.

jcarter3:
Created as an attempt to fix #324. Adds a new filter type of `date` which checks if the targeted field is _AFTER_ the current time + given offset, where offset is given in a `duration` compatible string.

Example:
```yaml
    filters:
      S3Bucket:
        - property: 'tag:expiration-date'
          type: date
          value: '-48h'
```
This would nuke an s3 bucket 2 days after the date specified in the `expiration-date` tag. Compatible date formats are:
* 1582735862 (seconds since epoch)
* 2006-01-02
* 2006/01/02
* 2006-01-02T15:04:05Z
* 2006-01-02T15:04:05.999999999Z07:00
* 2006-01-02T15:04:05Z07:00


jcarter3:
Is there any feedback for this?

svenwltr:
Hello @jcarter3. Sorry for the late response. I wanted to think about this and then forgot it.

* Is `value` always negative? If so, why not invert the value?
* Naming the type `date` sound a bit ambiguous to me. Especially when the `value` is some offset to the date.

jcarter3:
Not a problem, just wanted to make sure someone saw it.

* `value` is not always negative - it's an offset, though I'd say most use cases I can think of would have it be negative, so it might be more intuitive to invert it.
* `date` is very ambiguous, I don't like it, but just picked something easy so I could do a PoC. Maybe `dateAfter`, `dateOffset`, `dateCompare`... something?

Gowiem:
@svenwltr Sorry to do it, but pinging ya on this. Would be cool to see this ship!

svenwltr:
Hello @jcarter3 

Sorry for the late answer once again.

I guess `olderThan` or `dateOlderThan` would make it clear what it does. Also it would make it unambiguous whether the duration needs to be negative or not. What do you think?

Besides this, it looks fine to me.

jcarter3:
I think either of those name changes would work. I think the concept itself fits well with the project, but any name stands the risk of being pretty confusing - I can update the documentation to be as explicit as possible to help. I'll update it shortly

jcarter3:
I changed the name and updated the documentation to help clarify things. Also, I did make a minor switch to apply the offset given in `value` to the field time, _not_ the current time, which also makes it easier to understand.

svenwltr:
Thank you very much :)

wernerdiers:
Hi @jcarter3 
I was trying to use this feature but I cannot get it right.

Given this filter:
```
accounts:
  123456:
    filters:
      EC2Instance:
      - property: 'tag:expiration-date'
        type: dateOlderThan
        value: 2h
```
I'm getting the following error:
`Error: unknown type dateOlderThan`

I have gone through the README and the code at https://github.com/rebuy-de/aws-nuke/commit/f1eb408727fb0110caad65fc466c06d5bb5400cb and my filter config should work.

I'm using this Docker image version: 
`quay.io/rebuy/aws-nuke:v2.14.0 - aws-nuke version v2.14.0 - Thu Jan 30 14:05:49 UTC 2020 - d9bff8695c3a3e8f72050a76f14ce3f05c07b5c6`
and i tried also with `quay.io/rebuy/aws-nuke:v2.11.0` (taken from the README)



Could you please provide an example of the usage for this filter type? 

Thanks!

svenwltr:
Hello. This feature is not yet part of `v2.14`. It is only available with version `v2.15.0-beta.1`.

wernerdiers:
Thanks, it works flawlessly with that version. Sorry for the noice!!

k1m0n3:
Hello,
i'm trying to test this feature to delete EC2Volume older than 2 hours.

this code is used : 
accounts:
  # sandbox
  12345:
    filters:
      EC2Volume:
      - property: CreateTime
        type: dateOlderThan
        value: "2022-10-03T08:37:30.969Z"

But the filter is not applied, do you have any idea ? or working example ?

Thanks

Issue #430, opened 2019-11-12
nuke of  enabled CloudFormation stacks with "termination protected" enabled


youwalther65:
Would it be possible to add CloudFormation to feature-flag -> disable-deletion-protection

feature-flags:
  disable-deletion-protection:
    EC2Instance: true
    RDSInstance: true
    **CloudFormationStack: true**

Currently it seems it only supports EC2 and RDS
Thx in advance!

bjoernhaeuser:
@youwalther65 sure! Do you want to contribute this change?

jcarter3:
I added in support for this in a branch, but unfortunately did my other contribution out of `master` so will need to shuffle some stuff around to properly submit a PR

jcarter3:
Fixed the branches in my fork, submitted PR to support this via feature flag

jcarter3:
PR was merged

der-eismann:
Fixed by #481

PR #481, opened 2020-03-02
#430 - cloudformation support for disabling termination protection

jcarter3:
PR for supporting disabling termination protection when deleting cloudformation stacks (#430)

### DIFF
--- a/resources/cloudformation-stack.go
+++ b/resources/cloudformation-stack.go
@@ -2,13 +2,16 @@ package resources
 
 import (
 	"errors"
+	"strings"
+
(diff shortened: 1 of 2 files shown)


resources/cloudformation-stack.go
```diff
@@ -70,6 +91,7 @@ func (cfs *CloudFormationStack) removeWithAttempts(attempt int) error {
 }
 
 func (cfs *CloudFormationStack) doRemove() error {
+
```
der-eismann:
```suggestion

```

Issue #452, opened 2020-01-15
Disable Security Hub


gebailey:
I don't believe Security Hub gets disabled by aws-nuke.  Security Hub, if enabled, can incur costs associated with AWS Config rules and/or costs per finding.  Ideally, Security Hub would be disabled by aws-nuke in any regions that are processed.

RockAfeller2013:
also GuardDuty needs to be disabled. Ive tested and this script does not disable GuardDuty

ekristen:
This is implemented in the fork https://github.com/ekristen/aws-nuke 

If you have a chance, please check it out and let us know if you run into an issues by opening an issue over on the fork.

----

Please see the copy of the notice from the README about the deprecation of this project. Sven was kind enough to grant me access to help triage and close issues and pull requests that have already been addressed in the actively maintained fork. Some additional information is located in the [welcome issue](https://github.com/ekristen/aws-nuke/issues/276) for more information. 

> [!CAUTION]
> This repository for aws-nuke is no longer being actively maintained. We recommend users to switch to the actively maintained fork of this project at [ekristen/aws-nuke](https://github.com/ekristen/aws-nuke).
> We appreciate all the support and contributions we've received throughout the life of this project. We believe that the fork will continue to provide the functionality and support that you have come to expect from aws-nuke.
> Please note that this deprecation means we will not be addressing issues, accepting pull requests, or making future releases from this repository.
> Thank you for your understanding and support.

PR #484, opened 2020-03-05
Disable Security Hub

gebailey:
See https://github.com/rebuy-de/aws-nuke/issues/452

Add the ability to detect regions with Security Hub enabled and disable it if a
"hub" resource is returned.

### DIFF
--- a/resources/securityhub-hub.go
+++ b/resources/securityhub-hub.go
@@ -39,7 +39,7 @@ type Hub struct {
 
 func (hub *Hub) Properties() types.Properties {
 	properties := types.NewProperties()
-	properties.Set("HubArn", hub.id)
+	properties.Set("Arn", hub.id)
 	return properties
 }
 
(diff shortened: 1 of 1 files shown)


resources/securityhub-hub.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+type Hub struct {
+	svc *securityhub.SecurityHub
+	id  *string
+}
+
+func (hub *Hub) Properties() types.Properties {
+	properties := types.NewProperties()
+	properties.Set("HubArn", hub.id)
```
bjoernhaeuser:
```suggestion
	properties.Set("Arn", hub.id)
```

Otherwise we will have a lot of different `Arn` properties at some point :)

gebailey: Makes sense.  Additional commit pushed with this suggested fix.

PR #488, opened 2020-03-16
only list active stacksets; ignore deleted

tylersouthwick:


tylersouthwick:
Yes, not active and deleted

bjoernhaeuser:
And deleted means that they will be deleted or that they are deleted? What happens with these?

PR #489, opened 2020-03-23
Support for ApplicationAutoScaling - ScalableTargets

deanrock:
Resources such as dynamodb (sometimes?) leave behind `ScalableTargets` when deleted. Since there is a limit on how many of them can be on each account, we should support deleting them.

I couldn't find a programmatic way to find out all applicable `namespaces`, so I just hardcoded them.

### DIFF
--- a/resources/applicationautoscaling-scalabletargets.go
+++ b/resources/applicationautoscaling-scalabletargets.go
@@ -1,0 +1,85 @@
+package resources
+
+import (
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/applicationautoscaling"
+)
+
+func init() {
+	register("ApplicationAutoScalingScalableTarget", ListApplicationAutoScalingScalableTargets)
+}
+
+func ListApplicationAutoScalingScalableTargets(s *session.Session) ([]Resource, error) {
+	svc := applicationautoscaling.New(s)
+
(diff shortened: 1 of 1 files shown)


resources/applicationautoscaling-scalabletargets.go
```diff
@@ -0,0 +1,85 @@
...
+		"ecs",
+		"elasticmapreduce",
+		"ec2",
+		"appstream",
+		"dynamodb",
+		"rds",
+		"sagemaker",
+		"custom-resource",
+		"comprehend",
+		"lambda",
```
der-eismann: Please order them alphabetically.

der-eismann (dismissed):

resources/applicationautoscaling-scalabletargets.go
deanrock: 👍

svenwltr:
Hello,

is there a way to list available namespaces? Otherwise we have to maintain this list every time a new namespace comes up. Also, from looking at the docs, it looks like it is possible to define custom namespaces. Those would always be missing here, right?

svenwltr:
> Hello,
> 
> is there a way to list available namespaces? Otherwise we have to maintain this list every time a new namespace comes up. Also, from looking at the docs, it looks like it is possible to define custom namespaces. Those would always be missing here, right?

svenwltr:
> is there a way to list available namespaces? Otherwise we have to maintain this list every time a new namespace comes up. Also, from looking at the docs, it looks like it is possible to define custom namespaces. Those would always be missing here, right?

Closing because of missing response. Please address this and reopen, if there is still a desire for this.

PR #491, opened 2020-03-25
add firewall manager resources

tylersouthwick:


der-eismann (dismissed):

svenwltr:
Hello,

could you please add some properties? See https://github.com/rebuy-de/aws-nuke/blob/master/CONTRIBUTING.md#use-properties-instead-of-string-functions

bjoernhaeuser (dismissed):

resources/fms_policies.go
```diff
@@ -0,0 +1,69 @@
...
+	return err
+}
+
+func (f *FMSPolicy) String() string {
+	return *f.policy.PolicyId
+}
+
+func (f *FMSPolicy) Properties() types.Properties {
+	properties := types.NewProperties()
+	properties.Set("PolicyId", f.policy.PolicyId)
```
svenwltr, line 66:
```suggestion
	properties.Set("PolicyID", f.policy.PolicyID)
```

This does not follow the Golang naming conventions. Abbreviations should have consistent case. See [Guidelines](https://github.com/golang/go/wiki/CodeReviewComments#initialisms).

svenwltr (commented):
Sorry, this is the last nitpick :)

resources/fms_policies.go
tylersouthwick: we could set the property as `PolicyID` but, the field name from the aws sdk is `PolicyId`

bjoernhaeuser: I've updated this quickly. We try to stick with golang names as far as we can. We may have missed some though.

bjoernhaeuser (approved):

PR #499, opened 2020-04-28
Generate mocks before building

der-eismann:
After a [hint on the mailing list](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/aws-nuke/KmmMHDwf2oc/5ubmgoC3AQAJ) we noticed that the default `make` command doesn't work anymore because the mocks are missing. This should fix it.

### DIFF
--- a/golang.mk
+++ b/golang.mk
@@ -35,7 +35,7 @@ vendor: go.mod go.sum
 format:
 	gofmt -s -w $(GOFILES)
 
-vet: vendor
+vet: go_generate vendor
 	go vet $(GOPKGS)
 
 lint:
@@ -45,13 +45,13 @@ go_generate:
 	rm -rvf mocks
 	go generate ./...
 
-test_packages: vendor
+test_packages: go_generate vendor
 	go test $(GOPKGS)
 
 test_format:
 	gofmt -s -l $(GOFILES)
 
-test: test_format go_generate vet lint test_packages
+test: test_format vet lint test_packages
 
 cov: go_generate
 	gocov test -v $(GOPKGS) \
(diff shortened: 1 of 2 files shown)


golang.mk
```diff
@@ -28,7 +28,7 @@ OUTPUT_LINK=$(NAME)$(shell go env GOEXE)
 
 default: build
 
-vendor: go.mod go.sum
+vendor: go_generate go.mod go.sum
```
svenwltr, line 31: And `go_generate` will work before vendoring?

der-eismann: Yes, it **has** to be before vendoring.

golang.mk
```diff
@@ -65,17 +65,17 @@ _build: vendor
...
 compress: _build
 	tar czf dist/$(OUTPUT_FILE).tar.gz dist/$(OUTPUT_FILE)
 
-xc:
+xc: go_generate
 	GOOS=linux GOARCH=amd64 make compress
 	GOOS=darwin GOARCH=amd64 make compress
 
-install: vendor test
+install: test vendor
```
svenwltr, line 78: Well, that's weird now. A test won't work without vendoring before. Either the target naming is wrong now, or I won't work on a clean machine.

der-eismann: After investigating (because it's running fine on a clean machine) I noticed that `test` is running `test_packages` which itself is running `vendor`, so it's actually not used in the install step.

bjoernhaeuser (dismissed):

svenwltr (dismissed):

bjoernhaeuser (dismissed):

bjoernhaeuser (approved):

svenwltr (approved):

Issue #470, opened 2020-02-09
Cannot Delete IAM User with SSH Keys


mkhaled93:
Hi,

I am trying to delete some IAM users, Do we have any options to force the deletion rather than deleting the login profile, policy attachments and access keys first?

Because I can't find the resource SSH Keys for the Code Commit. So getting this error while trying deleting the IAM user with SSH keys attached: 
"ERRO[0046] DeleteConflict: Cannot delete entity, must remove referenced objects first"

Thanks

svenwltr:
Hello @mkhaled93. It would be possible to enable force deletion via feature-flags, if the AWS API allows it. Alternatively a resource for Code Commit Keys could be added.

wes-novack:
This was fixed via https://github.com/rebuy-de/aws-nuke/pull/533 , this issue should be good to close now.

PR #533, opened 2020-07-31
Delete IAM User SSH Public Keys

zerbitx:
We forked this for internal use, but noticed [this issu](https://github.com/rebuy-de/aws-nuke/issues/470).


svenwltr:
Thank you!

PR #538, opened 2020-08-27
Add API Gateway v2 APIs as a resource

der-eismann:


resources/apigatewayv2-apis.go
```diff
@@ -0,0 +1,60 @@
...
+func init() {
+	register("APIGatewayV2API", ListAPIGatewayV2APIs)
+}
+
+func ListAPIGatewayV2APIs(sess *session.Session) ([]Resource, error) {
+	svc := apigatewayv2.New(sess)
+	resources := []Resource{}
+
+	params := &apigatewayv2.GetApisInput{
+		MaxResults: aws.String("100"),
```
der-eismann, line 23: I have absolutely no idea why one would use a string for such an option
https://docs.aws.amazon.com/sdk-for-go/api/service/apigatewayv2/#GetApisInput

svenwltr: My guess is that something with API generation went wrong and that cannot get reverted without breaking other peoples stuff.

resources/apigatewayv2-apis.go
```diff
@@ -0,0 +1,60 @@
...
+func (f *APIGatewayV2API) Remove() error {
+
+	_, err := f.svc.DeleteApi(&apigatewayv2.DeleteApiInput{
+		ApiId: f.v2APIID,
+	})
+
+	return err
+}
+
+func (f *APIGatewayV2API) String() string {
```
svenwltr, line 58: Could you also add the `Properties()` thingy?

svenwltr (approved):

PR #548, opened 2020-08-31
Build binaries for Windows & ARM64

der-eismann:
@rebuy-de/prp-aws-nuke What's your opinion on this? 
1. Remove unpacked binaries in the `make compress` step to clean up our [release assets](https://github.com/rebuy-de/aws-nuke/releases/tag/v2.15.0-beta.1)
2. Build binaries for Windows and ARM64/ARMv7 linux
3. Remove `dist` path from archives when unpacking

### DIFF
--- a/golang.mk
+++ b/golang.mk
@@ -70,10 +70,13 @@ build: go_generate _build
 
 compress: _build
 	tar czf dist/$(OUTPUT_FILE).tar.gz dist/$(OUTPUT_FILE)
+	rm -f dist/$(OUTPUT_FILE)
 
 xc: go_generate
 	GOOS=linux GOARCH=amd64 make compress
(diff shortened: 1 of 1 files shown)


svenwltr (commented):
Personally I like having a plain binary for downloading, because I do not have to deal with unpacking and searching what directory structure this particular archive has (generally speaking).

der-eismann:
That's why I removed the directory structure so we only have the binary after unpacking. But having them both feels a bit superfluous.

svenwltr:
> That's why I removed the directory structure so we only have the binary after unpacking. But having them both feels a bit superfluous.

My point is more that every project has another directory structure in the archive. But anyways I am fine with skipping the binary.

bjoernhaeuser (approved):

PR #557, opened 2020-09-11
Enable CLICOLOR and CLICOLOR_FORCE env variables

icholy:
This allows forcing color output as described in https://bixense.com/clicolors/ . This is useful when using aws-nuke in CI jobs.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -31,6 +31,9 @@ func NewRootCommand() *cobra.Command {
 		if verbose {
 			log.SetLevel(log.DebugLevel)
 		}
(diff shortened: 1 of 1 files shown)


der-eismann:
I don't think we want to enable this by default, as this produces some hardly readable output on Jenkins if you are missing the plugins. Maybe we can add a flag. What's your opinion @rebuy-de/prp-aws-nuke ?

icholy:
@der-eismann The default behaviour is unchanged. It will only force output if the `CLICOLOR_FORCE=1` is set.

svenwltr (approved):

der-eismann (approved):

PR #576, opened 2020-11-26
Add EC2EgressOnlyInternetGateway resource

kevinkupski:
While nuking our account, I noticed that EC2 Egress Only Internet Gateways are not deleted and that the resource is not implemented in _aws-nuke_ yet.

This PR adds the _EC2EgressOnlyInternetGateway_ resource. It looks like an attachment resource is not necessary since the AWS API does not allow to delete an attachment, only the gateway itself.

Hint regarding Pagination: According to the [DescribeEgressOnlyInternetGateways](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeEgressOnlyInternetGateways.html) documentation, a  _MaxResults_ of up to 255 is supported.

### DIFF
--- a/resources/ec2-egress-only-internet-gateways.go
+++ b/resources/ec2-egress-only-internet-gateways.go
@@ -1,0 +1,72 @@
+package resources
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/rebuy-de/aws-nuke/pkg/types"
+)
+
+type EC2EgressOnlyInternetGateway struct {
+	svc *ec2.EC2
+	igw *ec2.EgressOnlyInternetGateway
+}
+
+func init() {
+	register("EC2EgressOnlyInternetGateway", ListEC2EgressOnlyInternetGateways)
+}
+
+func ListEC2EgressOnlyInternetGateways(sess *session.Session) ([]Resource, error) {
+	svc := ec2.New(sess)
+	resources := make([]Resource, 0)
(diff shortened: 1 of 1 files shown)


svenwltr (dismissed):

resources/ec2-egress-only-internet-gateways.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func init() {
+	register("EC2EgressOnlyInternetGateway", ListEC2EgressOnlyInternetGateways)
+}
+
+func ListEC2EgressOnlyInternetGateways(sess *session.Session) ([]Resource, error) {
+	svc := ec2.New(sess)
+
+	resp, err := svc.DescribeEgressOnlyInternetGateways(nil)
```
bjoernhaeuser, line 21: DescribeEgressOnlyInternetGateways could return more than one page. I think this i missing the token/nextToken handling like implemented here: https://github.com/rebuy-de/aws-nuke/blob/38b4ff979f1962f67caeeca011a34351ad640d33/resources/eks-fargate.go#L31-L47

kevinscholz: Thank you for the super fast feedback. I adjusted the PR accordingly. I was blindly using the structure from the existing _EC2InternetGateway_ resource. That one might be adjusted as well in a separate PR.

bjoernhaeuser: > I was blindly using the structure from the existing EC2InternetGateway resource. That one might be adjusted as well in a separate PR.

This would be awesome, thanks.

bjoernhaeuser (approved):

kevinscholz:
Thank you for the super fast feedback & merge. 👍 

I was just wondering: Do you already have a timeline for the next release?

PR #577, opened 2020-11-30
Add support for deleting cache parameter groups

kurtmc:


resources/elasticache-cacheparametergroups.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	_, err := i.svc.DeleteCacheParameterGroup(params)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
```
der-eismann, line 59: Please also add a `Properties()` function like [this one](https://github.com/rebuy-de/aws-nuke/blob/master/resources/ec2-instances.go#L109-L115). 
Rest LGTM

der-eismann (changes requested):

kurtmc:
@der-eismann I believe I have addressed your comment now.

PR #579, opened 2020-12-03
added lambda layers resource

mrkenkeller:
This adds deleting lambda layers to aws nuke.

As per the contribution guidelines I used `ListLayersPages` and added a `properties` function.

### DIFF
--- a/resources/lambda-layers.go
+++ b/resources/lambda-layers.go
@@ -1,0 +1,75 @@
+package resources
+
+import (
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/lambda"
+	"github.com/rebuy-de/aws-nuke/pkg/types"
+)
+
+type lambdaLayer struct {
+	svc       *lambda.Lambda
+	layerName *string
+	version   int64
+}
+
+func init() {
+	register("LambdaLayer", ListLambdaLayers)
+}
+
+func ListLambdaLayers(sess *session.Session) ([]Resource, error) {
+	svc := lambda.New(sess)
+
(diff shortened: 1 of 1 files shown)


resources/lambda-layers.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	resources := make([]Resource, 0)
+
+	for _, layer := range layers {
+
+		var i int64 = int64(1) // is there a better way to do this
+		for i <= *layer.LatestMatchingVersion.Version {
```
svenwltr, line 41: Shouldn't that be a generic for loop? Something like:

```go
for i := 1; i < *layer.LatestMatchingVersion.Version; i++ {
```

resources/lambda-layers.go
```diff
@@ -0,0 +1,75 @@
...
+			resources = append(resources, &lambdaLayer{
```
svenwltr, line 42: I don't understand this. Does lambda always have *all* versions from 1 to `LatestMatchingVersion`?

resources/lambda-layers.go
mrkenkeller: Sorry, old habits.  I'll fix and resubmit.

resources/lambda-layers.go
mrkenkeller: It did not throw an error when it tried to delete a version that did not exist (I tested that), but on reflection the right way is to use the list provided by `listLayerVersions`.  I'll update and resubmit.

svenwltr: Thanks. It is not really about errors, but about performance for cases where there is a high version number.

svenwltr (approved):

PR #592, opened 2021-01-18
Update sns

mrkenkeller:
Replaced `ListTopics` with `ListTopicsPages` to allow for delation of more than 100 sns topics.

Added `Properties` while I was in there.

Moved the `SNSTopic struct` struct to top of the document to be consistent with other resources.

### DIFF
--- a/resources/sns-topics.go
+++ b/resources/sns-topics.go
@@ -5,7 +5,14 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/sns"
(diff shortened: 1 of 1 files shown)


bjoernhaeuser (dismissed):
The git history looks a bit broken. Would be cool if you can fix it before merging :) Thanks!

mrkenkeller:
> The git history looks a bit broken. Would be cool if you can fix it before merging :) Thanks!

Is this better?

Sorry I was not sure how to fix it - let me know if it is not enough?

bjoernhaeuser (approved):

der-eismann (approved):

Issue #594, opened 2021-02-02
Documentation to Update Terminology


scottschreckengaust:
Consider migrating the `blacklist` terminology.  Some alternatives: `skiplist`, `excludelist`, or `ignorelist`

https://tools.ietf.org/id/draft-knodel-terminology-00.html#rfc.section.1.2

PR #595, opened 2021-02-02
use blocklist over blacklist

dfreudenberger:
addresses #594

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,16 +33,16 @@ To reduce the blast radius of accidents, there are some safety precautions:
 4. The Account Alias must not contain the string `prod`. This string is
    hardcoded and it is recommended to add it to every actual production account
    (eg `mycompany-production-ecr`).
-5. The config file contains a blacklist field. If the Account ID of the account
-   you want to nuke is part of this blacklist, *aws-nuke* will abort. It is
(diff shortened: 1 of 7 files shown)


svenwltr:
The problem is that this is a breaking change for all users. So either it causes a major release or we have to deprecate the old term and make both ones work for a while in parallel.

der-eismann:
Let's move from master to main in the same step and do a major release then

dfreudenberger:
Let me do the required changes in order to be backwards compatible. I think we shouldn't break compatibility if it's easy to avoid.

scottschreckengaust (dismissed):
Simple change, looks good.

scottschreckengaust:
> The problem is that this is a breaking change for all users. So either it causes a major release or we have to deprecate the old term and make both ones work for a while in parallel.

Good point to deprecate.

scottschreckengaust:
> Let's move from master to main in the same step and do a major release then

Bonus ;-)

svenwltr (dismissed):

der-eismann (dismissed):

svenwltr (approved):

der-eismann (approved):

PR #596, opened 2021-02-04
Adapt Readme & CI for main branch

der-eismann:
Follow up to #595

/cc @dfreudenberger

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,7 +2,7 @@ name: Golang CI
 
 on:
   push:
-    branches: [master]
+    branches: [main]
   pull_request:
     types: [opened, reopened, synchronize]
   schedule:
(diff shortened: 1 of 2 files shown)


svenwltr:
Do we break things like forked repos with this?

der-eismann:
I'd say it is the downstream's task to keep up with our changes, especially if it's such a rare occasion like a main branch change.

svenwltr (approved):

bjoernhaeuser:
> Do we break things like forked repos with this?

We do not. We only break everyone who is pulling "master". Is GitHub handling this for us? I do not expect it todo. We should probably delete the master branch and make every PR we have right now point to main afterwards.

der-eismann:
GitHub is actually handling this.
screenshot
But yeah, people pulling `master` from quay will have problems.

bjoernhaeuser (approved):

PR #608, opened 2021-03-09
Ignore errors when deleting Service Catalog Provisioned Products.

kurtmc:
I have run into an issue where aws-nuke will successfully delete the
    cloudformation stacks created by the service catalog but then fails to
    delete the service catalog as it is not able to find the cloudformation
    stacks to delete. I think adding the ignore errors flag will make
    aws-nuke more robust.


kurtmc:
@der-eismann any chance to get this reviewed?

svenwltr (approved):

bjoernhaeuser (approved):

PR #612, opened 2021-03-22
Add SageMaker studio resources

kurtmc:


kurtmc:
@der-eismann sorry to be annoying but as I am using this project more and more I can see how important it is that I can get changes merged in a reasonable timeframe. Perhaps we need to add some sort of integration tests. Also I can see that my PR has a failed check but its not related to my changes.

resources/sagemaker-apps.go
```diff
@@ -0,0 +1,90 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/sagemaker"
+	"github.com/rebuy-de/aws-nuke/pkg/types"
+)
+
+type SageMakerApp struct {
+	svc             *sagemaker.SageMaker
+	domainId        *string
```
svenwltr, line 14:
```suggestion
	domainID        *string
```

This does not follow the Golang naming conventions. Abbreviations should have consistent case. See [Guidelines](https://github.com/golang/go/wiki/CodeReviewComments#initialisms).

I know that the AWS-SDK does not follow that rule and some other names slippend through with an inconsistent naming, but we need to move towards a consistent naming.

svenwltr (commented):
The only nit pick on this PR is the naming of some variables and properties.

> sorry to be annoying but as I am using this project more and more I can see how important it is that I can get changes merged in a reasonable timeframe.

We are sorry, but we currently do not have much time to spare for aws-nuke.

> sorry to be annoying but as I am using this project more and more I can see how important it is that I can get changes merged in a reasonable timeframe. 

We tried working on that in the past, but the sheer amount of AWS resources makes this very hard and time-consuming.

> Also I can see that my PR has a failed check but its not related to my changes.

I just fixed that. It should not fail after a rebase on main.

PR #624, opened 2021-03-29
Add AppMesh resources

andskli:
Added the following AppMesh resources:
- Mesh
- Route
- VirtualGateway
- VirtualNode
- VirtualRouter
- VirtualService

### DIFF
--- a/resources/appmesh-mesh.go
+++ b/resources/appmesh-mesh.go
@@ -1,0 +1,64 @@
+package resources
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/appmesh"
+	"github.com/rebuy-de/aws-nuke/pkg/types"
+)
+
+type AppMeshMesh struct {
+	svc      *appmesh.AppMesh
+	meshName *string
+}
+
+func init() {
+	register("AppMeshMesh", ListAppMeshMeshes)
+}
+
+func ListAppMeshMeshes(sess *session.Session) ([]Resource, error) {
+	svc := appmesh.New(sess)
+	resources := []Resource{}
+
+	params := &appmesh.ListMeshesInput{
+		Limit: aws.Int64(100),
+	}
+
+	for {
+		output, err := svc.ListMeshes(params)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, item := range output.Meshes {
+			resources = append(resources, &AppMeshMesh{
+				svc:      svc,
+				meshName: item.MeshName,
+			})
+		}
+
+		if output.NextToken == nil {
+			break
+		}
+
+		params.NextToken = output.NextToken
+	}
+
+	return resources, nil
+}
+
+func (f *AppMeshMesh) Remove() error {
+	_, err := f.svc.DeleteMesh(&appmesh.DeleteMeshInput{
+		MeshName: f.meshName,
+	})
+
+	return err
+}
+
+func (f *AppMeshMesh) Properties() types.Properties {
+	properties := types.NewProperties()
+	properties.
+		Set("MeshName", f.meshName)
+
+	return properties
+}
--- a/resources/appmesh-virtualgateway.go
+++ b/resources/appmesh-virtualgateway.go
@@ -1,0 +1,85 @@
+package resources
+
+import (
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/appmesh"
+	"github.com/rebuy-de/aws-nuke/pkg/types"
+)
+
+type AppMeshVirtualGateway struct {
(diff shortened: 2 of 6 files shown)


bjoernhaeuser:
Hi @andskli,

could you please rebase your changes to latest changes in main? The build failures are fixed there. Thank you.

andskli:
@bjoernhaeuser it appears that did not change anything. :(

bjoernhaeuser:
> @bjoernhaeuser it appears that did not change anything. :(

Yes, we will investigate that and ignore for now :)

resources/appmesh-mesh.go
```diff
@@ -0,0 +1,64 @@
+package resources
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/appmesh"
+	"github.com/rebuy-de/aws-nuke/pkg/types"
+)
+
+type AppMeshMesh struct {
```
bjoernhaeuser, line 10: I am not sure about the naming here, I think `AppMesh` should be enough.

resources/appmesh-mesh.go
```diff
@@ -0,0 +1,64 @@
...
+		MeshName: f.meshName,
+	})
+
+	return err
+}
+
+func (f *AppMeshMesh) Properties() types.Properties {
+	properties := types.NewProperties()
+	properties.
+		Set("MeshName", f.meshName)
```
bjoernhaeuser, line 61: Normally we would say "Name" is enough, but this would allow to have the same filter on all "sub"resources of a AppMesh. @svenwltr @der-eismann whats your opinion here?

With "MeshName" all sub resources could be filtered the same way, with having Name here we would stay consistent with other resources, but the filter would look different.

resources/appmesh-virtualgateway.go
```diff
@@ -0,0 +1,85 @@
...
+	})
+
+	return err
+}
+
+func (f *AppMeshVirtualGateway) Properties() types.Properties {
+	properties := types.NewProperties()
+	properties.
+		Set("MeshName", f.meshName).
+		Set("VirtualGatewayName", f.virtualGatewayName)
```
bjoernhaeuser, line 82:
```suggestion
		Set("Name", f.virtualGatewayName)
```

resources/appmesh-virtualnode.go
```diff
@@ -0,0 +1,85 @@
...
+	})
+
+	return err
+}
+
+func (f *AppMeshVirtualNode) Properties() types.Properties {
+	properties := types.NewProperties()
+	properties.
+		Set("MeshName", f.meshName).
+		Set("VirtualNodeName", f.virtualNodeName)
```
bjoernhaeuser, line 82:
```suggestion
		Set("Name", f.virtualNodeName)
```

resources/appmesh-virtualrouter.go
```diff
@@ -0,0 +1,85 @@
...
+	})
+
+	return err
+}
+
+func (f *AppMeshVirtualRouter) Properties() types.Properties {
+	properties := types.NewProperties()
+	properties.
+		Set("MeshName", f.meshName).
+		Set("VirtualRouterName", f.virtualRouterName)
```
bjoernhaeuser, line 82:
```suggestion
		Set("Name", f.virtualRouterName)
```

resources/appmesh-virtualservice.go
```diff
@@ -0,0 +1,85 @@
...
+	})
+
+	return err
+}
+
+func (f *AppMeshVirtualService) Properties() types.Properties {
+	properties := types.NewProperties()
+	properties.
+		Set("MeshName", f.meshName).
+		Set("VirtualServiceName", f.virtualServiceName)
```
bjoernhaeuser, line 82:
```suggestion
		Set("Name", f.virtualServiceName)
```

bjoernhaeuser (commented):
The Name suggestions are for readability, in the output the resource is prefixed, no need to repeat the name again in the properties.

resources/appmesh-mesh.go
svenwltr: I'd say it should be consistent within a AWS-Service.

resources/appmesh-mesh.go
andskli: I opted for this given that AppMesh is the service name and Mesh is an API resource within the service. Similar to `APIGatewayV2API` ([ref](https://github.com/rebuy-de/aws-nuke/blob/main/resources/apigatewayv2-apis.go#L10)) is not `APIGatewayV2` named only.

I don't have a strong opinion, and are happy to change if you feel `AppMesh` is the appropriate name.

bjoernhaeuser: Valid point, let's keep it like this.

bjoernhaeuser (approved):

Issue #440, opened 2019-12-16
GovCloud support


ericfrederich:
This tool currently does not work with GovCloud.
The only region I have specified in my config is `us-gov-west-1`.  I get the following error:

`Error: failed get caller identity: InvalidClientTokenId: The security token included in the request is invalid.`

I've been working with GovCloud enough to know that this typically means it's using GovCloud credentials against the regular commercial AWS endpoints.  I tried exporting `AWS_DEFAULT_REGION` to `us-gov-west-1` in hopes that maybe the underlying golang SDK would pick up on this and use the proper endpoints but this did not work.

kylelaker:
If you want to change the source and recompile you can make the following changges in `pkg/awsutil/session.go`:

- Replace `endpoints.UsEast1RegionID` with `endpoints.UsGovWest1RegionID`
- Replace  `endpoints.AwsPartitionID` with `endpoints.AwsUsGovPartitionID`

Hopefully when #411 is resolved GovCloud support will work too.

kylelaker:
Another thing that might be worth mentioning is that since the GovCloud Administrator user is just a regular IAM user, aws-nuke will try to delete it as well as the Administrators group attachment (and the AdminPolicy policy attachment to that group). The API seems to happily allow these resources (and Login Profiles and Access Keys) to be deleted. You will want to ensure that these resources are excluded when running aws-nuke or that you have your GovCloud root access keys (though it is not necessarily typical to have them).

aghassemlouei:
Native support for GovCloud regions would be immensely beneficial as the error is the same when pointing to `us-gov-east-1` as well.

chrisgilmerproj:
@ericfrederich @kylelaker @aghassemlouei - I have a pull request open which ought to address this for you. At least you could pull the code from my branch and compile it if that helps.

lohman:
Noting that even with the above merge, this remains a problem in 2.17.0

yuvalyacoby:
running with `--default-region us-gov-west-1` worked for me 🎉

lohman:
Yep, that's the trick.  It requires that we specify that parameter even if the config has been updated with that region.  So maybe the dev could eliminate the need for that --default-region statement?

PR #647, opened 2021-06-02
Enable us-gov-cloud by setting the default partition correctly

chrisgilmerproj:
This addresses https://github.com/rebuy-de/aws-nuke/issues/440

The approach here is to allow folks to set the flag `--default-region`. If it is a known AWS region the correct partition is used. If it is not a known AWS region then the user is required to add endpoints into the config, which was the original default behavior. The idea is for this to augment but not replace or modify any existing functionality.

The constants being used can be found in the documentation here: https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/#pkg-constants. More configuration can be done for other regions and partitions but I leave that to others to implement.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"sort"
 
+	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/rebuy-de/aws-nuke/pkg/awsutil"
 	"github.com/rebuy-de/aws-nuke/pkg/config"
 	"github.com/rebuy-de/aws-nuke/resources"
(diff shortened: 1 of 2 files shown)


estebane-frb:
@chrisgilmerproj thanks for your reply. Any idea if there are plans to merge this to main and include in a future release?

chrisgilmerproj:
> @chrisgilmerproj thanks for your reply. Unfortunately this still did not work. My only workaround to get it working was to do as you suggested a while back and hardcode the DefaultRegionID and DefaultAWSPartitionID :(.

I have no idea when there will be a release. I assume you need to talk to @svenwltr .

estebane-frb:
@chrisgilmerproj, @svenwltr  I'm afraid that even with this merge, something is still not right - I get InvalidToken exception. I believe this is due to the endpoints in the pkg/awsutil/session.go. I took a fork and modified the DefaultRegionID and DefaultAWSPartitionID to endpoints.UsGovWest1RegionID and endpoints.AwsUsGovPartitionID, respectively, and this got rid of the error.

Unfortunately my employer does not permit me to submit PRs in OSS :( . I think if we add some condition like we did in cmd/root.go then this will resolve invalid token issue.

lohman:
Confirming it still doesn't work in 2.17.0.

svenwltr:
Hello. Unfortunately we cannot fix this, because we do not have access to the us-gov-cloud. So someone would have to provide a patch.

lohman:
Actually, I found last night that you're pretty close!  If I use the --region on the command line, much of it works.  It just doesn't properly utilize that region when I use it in the config.

I am no coder and therefore cannot submit a patch.  I can,  however, perform testing for you and provide logs.  It may be slow, but if you wanted to collaborate, I'd be happy to learn how to build the project and then test various things and provide logs so that perhaps we can figure it out together?

chrisgilmerproj:
Last time I used this my command looked like:

```
aws-nuke \
    --access-key-id "${AWS_ACCESS_KEY_ID}" \
    --secret-access-key "${AWS_SECRET_ACCESS_KEY}" \
    --session-token "${AWS_SESSION_TOKEN}" \
    --config "${PWD}"/nuke-config.yml \
    --default-region=us-gov-east-1 \
    --force \
    --force-sleep 3
```

The config looked exactly like what I had posted above:

```yaml
regions:
- us-gov-east-1
- us-gov-west-1
- global

account-blocklist:
- "999999999999" # fake as it is a required parameter

accounts:
  "123456789101": 
... # more config below
```

Unfortunately I'm not in a position to test this anymore so I can't provide any more help. If that changes I'll be happy to come back around and leave notes about what I find.

chrisgilmerproj:
@yuvalyacoby noted that the missing instruction for most folks is `--default-region us-gov-west-1`.

https://github.com/rebuy-de/aws-nuke/issues/440#issuecomment-1108685981

PR #662, opened 2021-07-27
Add tags to IAM resources

njtrettel:
Working on a use-case that requires us to filter IAM Policies by tags, which is not yet possible in aws-nuke.

This turned into a bigger change than originally expected. The reason is, the IAM `List*` APIs are not returning tags. The `Get*` APIs are, however. This was not obvious at first because the responses of the List and Get APIs are of the same _type_, which includes a `Tags` property. But alas, after testing I confirmed that the only the Gets in fact return them.

### The goal
A few improvements to the IAM resource filtering capabilities:
1. filter IAM policies by their tags
2. filter IAM role policies (inline policies) by their role's tags
   - _technically this was already implemented but it was broken because the code assumed the `List*` API returned the Tags (see above)_
3. filter IAM role policy attachments (managed policies) by their role's tags

### The solution

Needed to create two new shared functions: `GetIAMPolicy` and `GetIAMRole` to actually perform the `Get*` API call. I put these in `iam-roles.go` and `iam-policies.go` respectively, but I'm open to a more common location like a util.

After listing each of the mentioned resources, immediately make a Get call to grab the entire resource. Then, use extract properties from that instead of the object returned in the List API.

### DIFF
--- a/resources/iam-policies.go
+++ b/resources/iam-policies.go
@@ -38,7 +38,12 @@ func ListIAMPolicies(sess *session.Session) ([]Resource, error) {
 
 	err := svc.ListPoliciesPages(params,
 		func(page *iam.ListPoliciesOutput, lastPage bool) bool {
-			for _, policy := range page.Policies {
+			for _, listedPolicy := range page.Policies {
+				policy, err := GetIAMPolicy(svc, listedPolicy.Arn)
+				if err != nil {
+					logrus.Errorf("Failed to get listed policy %s: %v", *listedPolicy.PolicyName, err)
+					break
+				}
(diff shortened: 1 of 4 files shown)


der-eismann (dismissed):

njtrettel:
putting a WIP on this because it isn't working

the list policy API calls do not return tags, whether it's the go SDK or the CLI.

the get policy APIs do, however, so i'll need to rework this to call an extra `GetPolicy` API call and get the tags from there instead

svenwltr (approved):

bjoernhaeuser (approved):

PR #669, opened 2021-08-18
Add RDS proxy deletion support

elaygl:
Add support to delete RDS proxy resources in the AWS account.

### DIFF
--- a/resources/rds-proxies.go
+++ b/resources/rds-proxies.go
@@ -64,7 +64,7 @@ func (i *RDSProxy) String() string {
 
 func (i *RDSProxy) Properties() types.Properties {
 	properties := types.NewProperties()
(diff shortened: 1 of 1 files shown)


resources/rds-proxies.go
```diff
@@ -0,0 +1,74 @@
...
+	return nil
+}
+
+func (i *RDSProxy) String() string {
+	return i.id
+}
+
+func (i *RDSProxy) Properties() types.Properties {
+	properties := types.NewProperties()
+	properties.Set("Identifier", i.id)
```
svenwltr, line 67:
```suggestion
	properties.Set("ProxyName", i.id)
```

I would call it ProxyName, because that is the name of the field in the SDK.

PR #678, opened 2021-09-08
Add `CreationTime` to elasticache replication groups

rbroemeling:
Add the ability to easily filter for old or expired elasticache replication groups by adding the `CreationTime` property to them. This allows `aws-nuke` to easily clear out all "old" resources while leaving recent ones intact.

Note that this functionality requires the updated `aws-sdk-go` library version, as was PR'ed in #677.

### DIFF
--- a/resources/elasticache-replicationgroups.go
+++ b/resources/elasticache-replicationgroups.go
@@ -1,14 +1,17 @@
 package resources
 
 import (
+	"time"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/elasticache"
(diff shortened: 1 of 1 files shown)


der-eismann:
Since all your PRs are pretty small I would ask you to bundle them into one PR next time.

rbroemeling:
If you would like me to rejig this so that the `ElasticacheReplicationGroup` struct only pulls in the field necessary (`ReplicationGroupCreateTime`), I am happy to do so. I was following the way that the cloudformation stack does it (see #676), but in other places it is handled differently as well.

Between the options of "pulling in the entire metadata struct from AWS" and "pulling in only specific fields from the metadata struct from AWS", I think that I like pulling in only specific fields better, but I'm happy to be guided by the standard of `aws-nuke` here as well.

Let me know!

PR #680, opened 2021-09-09
Add `CreatedAt` property to EKS nodegroups

rbroemeling:
Add the ability to easily filter for old or expired EKS nodegroups by adding the `CreatedAt` property to them. This allows `aws-nuke` to easily clear out all "old" resources while leaving recent ones intact.

### DIFF
--- a/resources/eks-nodegroups.go
+++ b/resources/eks-nodegroups.go
@@ -2,6 +2,8 @@ package resources
 
 import (
 	"fmt"
+	"time"
+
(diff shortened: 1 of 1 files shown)


rbroemeling:
Same proviso applies to this as was written in #678 -- I am happy to update this PR to only pull in the specific field necessary, rather than the entire metadata struct from AWS, if that is what would be preferred.

der-eismann (approved):

bjoernhaeuser (approved):

PR #713, opened 2021-11-20
Logging tweaks

ekristen:
This adds an improvement around logging so that all logrus levels are supported and so that the log level can be changed from info to warn or error as an example.

`--verbose` is still there and will work, but logs a warn deprecated message informing the user they should use `--log-level=debug` instead.

The new option is `--log-level` and supports `trace,debug,info,warn,error`.

Why this change? Now you can set `--log-level=error` and make the output in conjunction with --quiet only log the resources and errors it encounters along the way.

Also adding `config.yaml` and `.envrc` to the .gitignore so for those doing development work `config.yaml` is automatically excluded and `.envrc` for [direnv](https://direnv.net) users such as myself.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@
 /dist
 /releases
 /mocks
+.envrc
+config.yaml
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"sort"
+	"strings"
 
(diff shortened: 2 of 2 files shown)


svenwltr:
Hello. I understand the desire of this PR, but I would really like to keep the `-v | --verbose` flags out of convenience (because the verbose flag exists at most of cli programs).

Do you see a way of not deprecating this?

ekristen:
Absolutely, simply remove the text that says it'll be deprecated and keep it so both paths work and the improvements remain.

PR #714, opened 2021-11-22
GoReleaser Support

ekristen:
## Overview

I think it would be worth considering merging this in to standardize builds, while the makefile is great, it doesn't cover release asset upload and docker builds, where goreleaser does.

This is a great tool that makes releasing golang based projects very simple.  This configuration is setup to build the same set of binaries already being built, but it also adds building docker images for each linux based build including arm, then publishing a manifest so a single docker pull will work on amd64, arm64, or arm32v7. This can easily be expanded to support other architectures.

## Differences

- Files within archive are just named `aws-nuke` instead of matching the name of the archive

## Testing 

If you'd like to take it for a test run you can simply clone my branch and run `gorelease release --rm-dist --snapshot` and it'll build everything locally for you.

## Releasing 

Performing a release then becomes as simple as cutting a tag and running the tool **OR** better yet using the GitHub Action.

```bash
git tag -a v2.17.0-rc1 -m 'Release v2.17.0-rc1'
git push --tags
goreleaser release --rm-dist
```

### GitHub Actions Support

**Important:** there is a github action that can be installed on GitHub Actions, if you go that route, then you simply need to only tag and push tags and GitHub Actions will take care of the rest via GitHub Actions.

## Example

You can see https://github.com/ekristen/aws-nuke/releases/tag/v2.17.0-ek.2 as an example

I did this branch because I needed a way to build releases of improvements I'm making since they haven't been merged into the main branch on the main repository. This coupled with a small bash script lets me sed replace the `debuy-de` with `ekristen` and publish under my fork, as the above link shows.

Try this from amd64, arm64 or arm32v7, and you should get the following back.

```bash
docker run -it --rm ghcr.io/ekristen/aws-nuke:2.17.0-ek.1 version
```

```
version:     2.17.0-ek.1
build date:  2021-11-22T22:43:03Z
scm hash:    885c7d50a103d5ffb9cee32b77a2f053ffd65e6d
environment: unknown
```

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -48,7 +48,7 @@ dockers:
       - "--label=org.opencontainers.image.title={{.ProjectName}}"
       - "--label=org.opencontainers.image.revision={{.FullCommit}}"
       - "--label=org.opencontainers.image.version={{.Version}}"
-      - "--label=org.opencontainers.image.source=https://{{ .ModulePath }}"
+      - "--label=org.opencontainers.image.source=https://github.com/debuy-de/aws-nuke"
       - "--platform=linux/amd64"
   - use: buildx
     goos: linux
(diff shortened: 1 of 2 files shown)


svenwltr:
Hello. Thanks for you submission. We do not really see a benefit of goreleaser over the existing build pipeline. Sure it is nice to just use an existing tool to get a project up faster, but _aws-nuke_ already has a working build pipeline so IMO there has to be a bit more than that.

ekristen:
Understood. Sure it does, but it's scope is limited. 

Goreleaser allows for the same outputs you already want but then has Broader support for building for other architecture by modifying a single config.

The example I did is doing multi architecture docker images and a manifest so every arch support is available via docker. 

It provides automatic change logs too. 

It also makes support for contributors that want to fork and test fixes and improvements much easier.

Issue #503, opened 2020-05-13
I keep getting a DependencyViolation errors!


john-morsley:
AWS Nuke throws up multiple errors:

```
DependencyViolation: The vpc ... has dependencies and cannot be deleted.
```

I can't see why. When I attempt to delete them via the AWS console, they delete without any problem.

Any ideas, or is this a bug!?

svenwltr:
Hello @john-morsley.

Sorry for the late response. It probably works through the UI, because the UI is doing a recursive delete. When using the API on the other hand AWS expects us to manually delete all dependent resources.

This means there must be a resource which depends on the VPC which is either not yet supported by aws-nuke or which is filtered by the config.

der-eismann:
Any news on this @john-morsley? Were you able to find a solution for your problem?

fyangd:
Bumping this thread, we recently started to observe this issue as well.  Looking in the AWS console, there doesn't appear to be other dependencies associated with the VPC.  Also, retrying with aws-nuke shows that no other dependencies exist,

```
[2021-04-22T16:08:31.056Z] us-west-2 - EC2InternetGateway - igw-0dadd35a683869d51 - [tag:Name: "default"] - filtered by config

[2021-04-22T16:08:31.056Z] us-west-2 - EC2Subnet - subnet-f4bfa78d - [DefaultForAz: "true"] - filtered by config

[2021-04-22T16:08:31.056Z] us-west-2 - EC2Subnet - subnet-2bec0676 - [DefaultForAz: "true"] - filtered by config

[2021-04-22T16:08:31.056Z] us-west-2 - EC2Subnet - subnet-d1e3e69a - [DefaultForAz: "true"] - filtered by config

[2021-04-22T16:08:31.057Z] us-west-2 - EC2Subnet - subnet-ea81e3c1 - [DefaultForAz: "true"] - filtered by config

[2021-04-22T16:08:31.057Z] us-west-2 - EC2RouteTable - rtb-0b6753bdae6283b8d - [] - filtered by config

[2021-04-22T16:08:31.057Z] us-west-2 - EC2RouteTable - rtb-86c8a0fd - [] - filtered by config

[2021-04-22T16:08:31.057Z] us-west-2 - EC2VPC - vpc-0898dcfe306f12f1e - [ID: "vpc-0898dcfe306f12f1e", IsDefault: "false",] - would remove

[...]

PR #720, opened 2021-12-15
Removed ENI status property to fix logic issue

sas-pemcne:
I have removed the Status property from the EC2NetworkInterface resource to avoid issues where the nuker thinks the resource has been deleted when it has not. 

According to the logic of the nuker: 
1. Resources are triggered for deletion
2. All resources are queried again
3. If the resource in the query matches a resource in the initial cache, the resource still exists, otherwise it has been successfully deleted

A problem arises in the matching logic because all properties of the resource must be identical for the match to be successful. This dictates that all properties must be immutable for the nuker to function properly. The status property of a EC2NetworkInterface would change from "in-use" to "available" (assuming it was attached) during the deletion process. Because of that, the resource could not be matched (step 3 above) and the nuker prematurely assumes the ENI has been deleted when the delete command had not been successfully issued. 

This might help some issues related to VPC dependency errors like #503 since any attached ENI will not be removed correctly. I also realize this will cause some breaking changes if others have this filtered in the config but I couldn't think of an easier way to resolve the issue without overhauling the matching logic.

### DIFF
--- a/resources/ec2-network-interfaces.go
+++ b/resources/ec2-network-interfaces.go
@@ -1,6 +1,9 @@
 package resources
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
@@ -37,13 +40,17 @@ func ListEC2NetworkInterfaces(sess *session.Session) ([]Resource, error) {
 }
 
 func (e *EC2NetworkInterface) Remove() error {
+
 	if e.eni.Attachment != nil {
 		_, err := e.svc.DetachNetworkInterface(&ec2.DetachNetworkInterfaceInput{
 			AttachmentId: e.eni.Attachment.AttachmentId,
 			Force:        aws.Bool(true),
 		})
 		if err != nil {
-			return err
+			expected := fmt.Sprintf("The interface attachment '%s' does not exist.", *e.eni.Attachment.AttachmentId)
+			if !strings.Contains(err.Error(), expected) {
+				return err
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


ekristen:
Woah, great find, I was just reviewing the code and agree with your assessment, I actually think this behavior is bad. Properties shouldn't have to match especially when resources have unique ids and those properties can change. 

This might be a bug in the core queue handling that needs to be addressed.

svenwltr:
I agree with @ekristen.

The original idea was to deprecate the `String()` function for resources and use `Properties()` only, but this was never pushed forward. Therefore most of the resources still have this. I think making the `String()` function not optional anymore and using this as the primary key would solve all bugs of this kind. Also it might make sense to rename it to `ID()` or similar in another step, but that's not necessary.

For this particular resource I suggest adding the String() function and return the `NetworkInterfaceId` there.

sas-pemcne:
Good catch with the String() comparison, I was so focused on the Properties that I completely missed that check in the Equals function. I added the String() and local testing looks good. 

A byproduct of actually having the ENI being deleted correctly is it was complaining about the attachment not existing when it tries to remove it and the Remove() would throw that error back. I'm guessing this is because the resource itself is not updated as the nuker progresses so it still thinks it has an attachment when it in fact does not. That or I was hitting a very coincidental timing issue during the detaching process. I removed the error return just for the detachment process which matches the same logic used in the EC2SecurityGroups where the pre-removal steps are performed and the output is not logged. This allows the ENI to be properly removed since it can actually make it down to the delete call.

resources/ec2-network-interfaces.go
```diff
@@ -37,14 +37,12 @@ func ListEC2NetworkInterfaces(sess *session.Session) ([]Resource, error) {
 }
 
 func (e *EC2NetworkInterface) Remove() error {
+
 	if e.eni.Attachment != nil {
-		_, err := e.svc.DetachNetworkInterface(&ec2.DetachNetworkInterfaceInput{
+		e.svc.DetachNetworkInterface(&ec2.DetachNetworkInterfaceInput{
```
svenwltr, line 42: Ignoring the error completely has the problem that it would assume the detachment worked properly on all other issues too (for example a connection issue).

Can we specifically check that the error indicates that the network interface is already gone?

sas-pemcne: I added a check for the specific error and ignore it, otherwise if it is unexpected it will get returned back.

PR #721, opened 2021-12-20
Add resource for EC2DefaultSecurityGroupRule

### LINKED ISSUE #410: Cannot delete VPC because of default security group having ingress/egress rules to other security groups
Hello,

In the following situation:

- vpc has a security group A
- vpc has also a default security group
- default security group have rules mentioning security group A

aws-nuke fails to delete all the resources.

Why ? That's because default security groups cannot be deleted in AWS. So the ingress or egress rule are not deleted. Thus when trying to delete security group A, AWS says it has dependencies and cannot be deleted.

Possible solution: make aws-nuke revoke all the rules in the default security group before trying to delete the security groups of the VPC.

Example default security group:

```json                              
{
    "SecurityGroups": [                                                                                                                                                                        
        {
            "Description": "default VPC security group",                                                                                                                                       
            "GroupName": "default",        
            "IpPermissions": [                                                                 
                {
                    "IpProtocol": "-1",                                                                                                                                                        
                    "IpRanges": [],                                                            
                    "Ipv6Ranges": [],
                    "PrefixListIds": [],
                    "UserIdGroupPairs": [
                        {                                                                                                                                                                      
                            "GroupId": "sg-042c737a0a360fb12",
                            "UserId": "725027129649"
                        },
                        {
                            "GroupId": "sg-0fb6fd9423ccf2387",
                            "UserId": "725027129649"
                        }
                    ]
                }
            ],
            "OwnerId": "725027129649",
            "GroupId": "sg-042c737a0a360fb12",
            "IpPermissionsEgress": [],
            "VpcId": "vpc-0bf9e729380fa85f0"
        }
    ]
}
```

VPC:

```json
{                                                                                             
    "Vpcs": [                                                                                  
        {                                                                                      
            "CidrBlock": "10.0.0.0/16",                                                        
            "DhcpOptionsId": "dopt-0f1ae70cd8426cf84",                             
            "State": "available",                                                              
            "VpcId": "vpc-0bf9e729380fa85f0",                                                  
            "OwnerId": "725027129649",                                                                                                                                                         
            "InstanceTenancy": "default",                                                      
            "CidrBlockAssociationSet": [                                                       
                {                                                                              
                    "AssociationId": "vpc-cidr-assoc-063bf1030a24b1963",           
                    "CidrBlock": "10.0.0.0/16", 
                    "CidrBlockState": {
                        "State": "associated"
                    }
                }
            ],
            "IsDefault": false,
            "Tags": [
                {
                    "Key": "kubernetes.io/cluster/cluster-2d3c-x4z74",
                    "Value": "owned"
                },
                {
                    "Key": "Name",
                    "Value": "cluster-2d3c-x4z74-vpc"
                }
            ]
        }
    ]
}
```

sas-pemcne:
I have added a resource to delete default security group rules. The default security group itself cannot be deleted and is filtered out but the rules can and should get deleted to prevent dependency issues where other security groups are referenced. This would fix #410. 

With the discussion on #410, I opted to just target the security group rules on the default security group alone since the EC2SecurityGroup removal function deletes the rules out of the others. This would provide a minimal configuration update to opt out.

### DIFF
--- a/resources/ec2-default-security-group-rules.go
+++ b/resources/ec2-default-security-group-rules.go
@@ -1,0 +1,108 @@
+package resources
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/rebuy-de/aws-nuke/pkg/types"
+)
+
+type EC2DefaultSecurityGroupRule struct {
+	svc      *ec2.EC2
+	id       *string
+	groupId  *string
+	isEgress *bool
+}
+
+func init() {
+	register("EC2DefaultSecurityGroupRule", ListEC2SecurityGroupRules)
+}
+
+func ListEC2SecurityGroupRules(sess *session.Session) ([]Resource, error) {
+	svc := ec2.New(sess)
+	resources := make([]Resource, 0)
+
+	sgFilters := []*ec2.Filter{
+		{
+			Name: aws.String("group-name"),
+			Values: []*string{
+				aws.String("default"),
+			},
+		},
+	}
+	groupIds := make([]*string, 0)
+	sgParams := &ec2.DescribeSecurityGroupsInput{Filters: sgFilters}
+	err := svc.DescribeSecurityGroupsPages(sgParams,
+		func(page *ec2.DescribeSecurityGroupsOutput, lastPage bool) bool {
+			for _, group := range page.SecurityGroups {
+				groupIds = append(groupIds, group.GroupId)
+			}
+			return !lastPage
+		})
+	if err != nil {
+		return nil, err
+	}
+
+	sgRuleFilters := []*ec2.Filter{
+		{
+			Name:   aws.String("group-id"),
+			Values: groupIds,
+		},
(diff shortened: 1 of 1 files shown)


sas-pemcne:
Just realized with the other discussion about the String() function being the unique identifier that this resource will also need to return the rule ID and not the parent security group rule. Let me update that first.

sas-pemcne:
Updated to use the security group rule id as the unique identifier. Output will be similar to the following: 

```
us-east-1 - EC2DefaultSecurityGroupRule - sgr-0e4569ceff05076dd - [SecurityGroupId: "sg-019a122f4c02f84b4"] - would remove
us-east-1 - EC2DefaultSecurityGroupRule - sgr-0d02a20a9bbf4b857 - [SecurityGroupId: "sg-019a122f4c02f84b4"] - would remove
us-east-1 - EC2DefaultSecurityGroupRule - sgr-04935aea59688b50e - [SecurityGroupId: "sg-019a122f4c02f84b4"] - would remove
us-east-1 - EC2DefaultSecurityGroupRule - sgr-0df81fead809245a2 - [SecurityGroupId: "sg-019a122f4c02f84b4"] - would remove
```

PR #729, opened 2022-01-18
adding ability to nuke macie from an AWS account

ekristen:
Macie will delete everything automatically if you disable it so we do not need to iterate over it's resources to nuke it, we can just disable it. If individual Macie resources need removal, additional resource types can be defined.


ekristen:
Thanks for the quick review and merge!

PR #731, opened 2022-01-20
feat: add missing tags for iam resources

mavogel:
based on https://github.com/rebuy-de/aws-nuke/pull/662 I added tags for the resources
- `iam-user`
- `iam-instance-profile`


mavogel:
TY @svenwltr :) when will be the next release?

PR #734, opened 2022-01-23
Add tags and properties to some resources

hv202x1:
I added properties and tags to some resources. Everything is sorted in small commits.

### DIFF
--- a/resources/autoscaling-groups.go
+++ b/resources/autoscaling-groups.go
@@ -4,6 +4,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
+	"github.com/rebuy-de/aws-nuke/pkg/types"
 )
 
 func init() {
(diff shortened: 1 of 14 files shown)


resources/ec2-dhcp-options.go
```diff
@@ -7,10 +7,9 @@ import (
 )
 
 type EC2DHCPOption struct {
-	svc        *ec2.EC2
-	id         *string
-	tags       []*ec2.Tag
-	defaultVPC bool
+	svc  *ec2.EC2
+	id   *string
+	tags []*ec2.Tag
```
der-eismann, line 12: Please keep the `defaultVPC` property. In general I'm unsure why you modified that file, I can't see any improvements.

der-eismann (commented):
What's your reasoning behind adding the `OwnerID` to so many resources?

resources/ec2-dhcp-options.go
hv202x1: Sorry this was an misstake, it should not be removed. Thanks for the catch.

For some resources it is possible to share them across accounts, for example, with VPC. The VPC then belongs to another account and can actually hopefully only be deleted from this account.

With the OwnerID it is now possible to filter these shared resources. This has the advantage, that there are fewer errors.

der-eismann (approved):

Issue #733, opened 2022-01-21
IAMInstanceProfile and IAMInstanceProfileRole don't support tags


hs-siso:
## Description

I am having an automated script to remove IAM resources based on a specific tag.

In the config file I have the following filtering applied:
```
        - property: 'tag:expiration_date'
          type: dateOlderThan
          value: 0
```
which works for `IAMRole`, `IAMPolicy`, and `IAMRolePolicyAttachment` but **does not work** for `IAMInstanceProfileRole` and `IAMInstanceProfile`.

Unfortunately these two IAM resources don't support filtering by tag.

The actual error:
```
global - IAMInstanceProfile - manual-destoy-env - would remove
global - IAMInstanceProfileRole - manual-destoy-env -> manual-destoy-env - would remove
```

There is an enforced dependency when trying to remove global IAM resources. Hence, removing an IAM role won't work if `IAMInstanceProfile` and `IAMInstanceProfile` haven't been removed previously. 

Can anyone help with this, please?

mavogel:
Hi @hs-siso, I recently added it for `IAMRole` in this PR: https://github.com/rebuy-de/aws-nuke/pull/731, however for `IAMInstanceProfileRole` it's still missing. I took the tf provider as a baseline, if I could find tags there: see [here](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_instance_profile).

Feel free to take my PR as a baseline and create one with the missing ones :)

hs-siso:
Hey @mavogel, I don't have write access to create branch and propose a fix for this though:
> You’re making changes in a project you don’t have write access to.

Will you be able to help sort this out?

hs-siso:
I have built from master and executed a test locally with the change that you've merged and am still observing that only the Name property is filled up. The Tags aren't.

Here's an example for a test environment (aws-nuke targeting `InstanceProfile` only):
```
global - IAMInstanceProfile - role-witty-fear-postgres-v10 - [Name: "role-witty-fear-postgres-v10"] - would remove
```

whereas with aws cli I can strictly observe that the Tags are there:
```
ᐅ aws iam get-instance-profile --instance-profile-name role-witty-fear-postgres-v10 --output json | jq '.InstanceProfile.Tags'
[
  {
    "Key": "owner",
    "Value": "someone"
  },
  {
    "Key": "cluster",
    "Value": "witty-fear-postgres-v10"
  },
  {
    "Key": "resource_owner",
[...]

PR #741, opened 2022-02-07
fix: instanceprofilerolebug

mrkenkeller:
When scanning the IAMInstanceProfileRole Nuke is currently displaying a lot of of data in addition and not just the the ProfileRole.  

This breaks the functionality, and old configs will not filter any of the IAMInstanceProfileRole. 
 This should be fixed ASAP.


mrkenkeller:
Looks like this was introduced in #733

der-eismann (approved):

Issue #745, opened 2022-02-21
Please support nuking MWAA


Nuru:
Please support nuking AWS Managed Workflows for Apache Airflow ([MWAA](https://aws.amazon.com/managed-workflows-for-apache-airflow/))

PR #737, opened 2022-01-28
prototype cloud-control-api support

svenwltr:
Experimental implementation of replacing resource discovery with Cloud Control API.

🔥 This is just a proof of concept and not ready for merging 🔥

Issue #692, opened 2021-10-01
leverage AWS Cloud Control API


PatMyron:
`ListResource` and `DeleteResource` from https://aws.amazon.com/blogs/aws/announcing-aws-cloud-control-api/ seem like an easier way to increase type coverage

svenwltr:
We are working on this in https://github.com/rebuy-de/aws-nuke/pull/753

der-eismann:
Closing because #753 was merged.

PR #753, opened 2022-03-04
add support for cloud control

svenwltr:
> Supersedes https://github.com/rebuy-de/aws-nuke/pull/737

Adds support for the Cloud Control API. Cloud Control API resources will behave like regular (now called "old-style") resources as seen in this example:

### Related Tickets:

* added resources: https://github.com/rebuy-de/aws-nuke/issues/745
* https://github.com/rebuy-de/aws-nuke/issues/692

```
eu-west-1 - AWS::EC2::VPC - vpc-08852bdafa9189b5c - [CidrBlock: "172.22.16.0/20", Identifier: "vpc-08852bdafa9189b5c", InstanceTenancy: "default", VpcId: "vpc-08852bdafa9189b5c"] - would remove
```

---

The output of `go run go run ./dev/list-cloudcontrol` looks like this:

image

---

### TODO

* [x] Properly handle properties (see `resources/cloudcontrol.go`). For example we need to convert `CidrBlockAssociations` and `Tags` into a format, that is usable by the filter config.
```json
{
  "VpcId": "vpc-456",
  "InstanceTenancy": "default",
  "CidrBlockAssociations": [
    "vpc-cidr-assoc-1234"
  ],
  "CidrBlock": "10.10.0.0/16",
  "Tags": [
    {
      "Value": "Kubernetes VPC",
      "Key": "Name"
    }
  ]
}
```
```
# Currently it would simply look like this:
eu-west-1 - AWS::EC2::VPC - vpc-456 - [CidrBlock: "10.10.0.0/16", CidrBlockAssociations: "[vpc-cidr-assoc-1234]", Identifier: "vpc-08852bdafa9189b5c", InstanceTenancy: "default", Tags: "[map[Key:Name Value:Kubernetes VPC]]", VpcId: "vpc-456"] - would remove
```
* [ ] ~Handle rate limiting~ I cannot reproduce the rate limit with the new approach to cloud control
* [x] Add actual removal
* [x] add documentation
* [ ] ~handle global services (eg S3) better: currently they get listed for every region~ I do not have a solution to this, yet.

### DIFF
--- a/resources/cloudcontrol.go
+++ b/resources/cloudcontrol.go
@@ -7,6 +7,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/cloudcontrolapi"
+	"github.com/google/uuid"
 	"github.com/pkg/errors"
 	"github.com/rebuy-de/aws-nuke/pkg/types"
 	"github.com/sirupsen/logrus"
@@ -23,6 +24,8 @@ func init() {
 	registerCloudControl("AWS::Timestream::Database")
 	registerCloudControl("AWS::Timestream::ScheduledQuery")
 	registerCloudControl("AWS::Timestream::Table")
+	registerCloudControl("AWS::Transfer::Workflow")
+	registerCloudControl("AWS::Synthetics::Canary")
 }
 
 func NewListCloudControlResource(typeName string) func(*session.Session) ([]Resource, error) {
@@ -69,10 +72,11 @@ func NewListCloudControlResource(typeName string) func(*session.Session) ([]Reso
 				}
 
 				resources = append(resources, &CloudControlResource{
-					svc:        svc,
-					typeName:   typeName,
-					identifier: identifier,
-					properties: properties,
+					svc:         svc,
+					clientToken: uuid.New().String(),
+					typeName:    typeName,
+					identifier:  identifier,
+					properties:  properties,
 				})
 			}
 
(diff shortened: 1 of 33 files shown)


resources/cloudcontrol.go
```diff
@@ -0,0 +1,160 @@
...
+	registerCloudControl("AWS::AppRunner::Service")
+	registerCloudControl("AWS::ApplicationInsights::Application")
+	registerCloudControl("AWS::Athena::DataCatalog")
+	registerCloudControl("AWS::Backup::Framework")
+	registerCloudControl("AWS::MWAA::Environment")
+	registerCloudControl("AWS::Synthetics::Canary")
+	registerCloudControl("AWS::Timestream::Database")
+	registerCloudControl("AWS::Timestream::ScheduledQuery")
+	registerCloudControl("AWS::Timestream::Table")
+	registerCloudControl("AWS::Transfer::Workflow")
```
der-eismann, line 35: I don't think it makes sense to keep a list, an advantage of CloudControl would be to immediately use a new resource after release because the APIs are the same. If we need to patch aws-nuke first it's not an advantage anymore. But I understand that we need it at the moment.

der-eismann (approved):

resources/cloudcontrol.go
svenwltr: So you do not agree with the reasoning I wrote in the comment above the list?

der-eismann: I said I do understand it, but we should try to get rid of it in the future.

svenwltr: IMO that will be only possible via a major release.

bjoernhaeuser (approved):
I like this a lot. With this we can get some feedback on the cloud-control part and then release a new major where the old-style resources are removed - we should consider automatic rewrite of the config then.

svenwltr:
> we should consider automatic rewrite of the config then.

Automatic rewrite from old-style to cloud control will be a lot of work, because we have to do the mapping for every resource and every property manually.

PR #772, opened 2022-04-07
Feature 771 S3 CreationDate property

### LINKED ISSUE #771: S3 CreationDate filter
I want to nuke S3 buckets and objects that are older than *x* hours.

I have the following config.
```yml
accounts:
  xxxxxxxxxxxx:
    filters:      # never nuke these resources:
      EC2Instance:
        - property: LaunchTime # EC2 instances that are younger x hours old
          type: dateOlderThan
          value: "100h"
      S3Bucket:
        - property: Name       # S3 buckets that are named "foo"
          value: "foo"
        - property: CreationDate # S3 buckets that are younger x seconds old
          type: dateOlderThan
          value: "100h"
      S3Object:
        - property: Bucket     # S3 objects that are in the bucket "foo"
          value: "foo"
        - property: CreationDate # S3 objects that are younger x seconds old
          type: dateOlderThan
          value: "100h"
```

The EC2Instance dateOlderThan filter works as expected, saving all EC2Instances that are younger than 100 hours.

The S3Bucket and S3Object dateOlderThan filter does not seem to work. For example, I made a bucket named `my-new-bucket-aws-nuke` on April 7, 2022, 16:57:35 (UTC+01:00), however aws-nuke outputs
```
eu-west-2 - S3Bucket - s3://my-new-bucket-aws-nuke - [Name: "my-new-bucket-aws-nuke"] - would remove
```

I see that buckets do have an accessible [CreationDate](https://docs.aws.amazon.com/AmazonS3/latest/API/API_Bucket.html). What am I missing?

jamesgeddes:
Adds the CreationDate to S3 buckets and objects so that one can filter S3 by how old it is. 

Resolves #771 

Thanks!

### DIFF
--- a/resources/s3-buckets.go
+++ b/resources/s3-buckets.go
@@ -140,6 +140,7 @@ func (e *S3Bucket) RemoveAllObjects() error {
 func (e *S3Bucket) Properties() types.Properties {
 	properties := types.NewProperties()
 	properties.Set("Name", e.name)
+	properties.Set("CreationDate", e.CreationDate())
 
 	for _, tag := range e.tags {
 		properties.SetTag(tag.Key, tag.Value)
--- a/resources/s3-buckets.go
+++ b/resources/s3-buckets.go
@@ -139,7 +139,7 @@ func (e *S3Bucket) RemoveAllObjects() error {
 
 func (e *S3Bucket) Properties() types.Properties {
 	properties := types.NewProperties()
-	properties.Set("Name", e.name)
+	properties.Set("Name", e.name).
 	properties.Set("CreationDate", e.CreationDate())
 
 	for _, tag := range e.tags {
--- a/resources/s3-buckets.go
+++ b/resources/s3-buckets.go
@@ -140,7 +140,7 @@ func (e *S3Bucket) RemoveAllObjects() error {
 func (e *S3Bucket) Properties() types.Properties {
 	properties := types.NewProperties()
 	properties.Set("Name", e.name).
-	properties.Set("CreationDate", e.CreationDate())
+	properties.Set("CreationDate", e.CreationDate)
 
 	for _, tag := range e.tags {
 		properties.SetTag(tag.Key, tag.Value)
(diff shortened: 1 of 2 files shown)


resources/s3-objects.go
```diff
@@ -102,7 +102,8 @@ func (e *S3Object) Properties() types.Properties {
 		Set("Bucket", e.bucket).
 		Set("Key", e.key).
 		Set("VersionID", e.versionID).
-		Set("IsLatest", e.latest)
+		Set("IsLatest", e.latest).
+		Set("CreationDate", e.CreationDate())
```
jamesgeddes, line 106: I'm fairly new to Go so was not sure if it should be `e.CreationDate()` or `e.CreationDate`. I would be grateful of a poke in the right direction 😄

resources/s3-buckets.go
```diff
@@ -140,7 +140,8 @@ func (e *S3Bucket) RemoveAllObjects() error {
 
 func (e *S3Bucket) Properties() types.Properties {
 	properties := types.NewProperties()
-	properties.Set("Name", e.name)
+	properties.Set("Name", e.name).
+	properties.Set("CreationDate", e.CreationDate())
```
jamesgeddes, line 144: I'm fairly new to Go so was not sure if it should be `e.CreationDate()` or `e.CreationDate`. I would be grateful of a poke in the right direction 😄

resources/s3-objects.go
jamesgeddes: According to the [docs](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/s3-example-basic-bucket-operations.html#:~:text=creation%20date%20of%20each%20bucket), it would appear that I should have gone with `e.CreationDate` - will update.

resources/s3-buckets.go
jamesgeddes: Same as [above](https://github.com/rebuy-de/aws-nuke/pull/772#discussion_r845460140)

PR #821, opened 2022-06-25
fix: Correctly filter policy attachments in aws-us-gov

kazimierzbudzyk:
In gov regins service policies have different prefix, and are formatted like `arn:aws-us-gov:iam::aws:policy/aws-service-role/AWSElasticLoadBalancingServiceRolePolicy`. Before this change running aws-nuke against gov accounts results in the following error:
```
level=error msg="UnmodifiableEntity: Cannot perform the operation on the protected role 'AWSServiceRoleForElasticLoadBalancing' - this role is only modifiable by AWS\n\tstatus code: 400
```

### DIFF
--- a/resources/iam-role-policy-attachments.go
+++ b/resources/iam-role-policy-attachments.go
@@ -80,7 +80,7 @@ func ListIAMRolePolicyAttachments(sess *session.Session) ([]Resource, error) {
 }
 
 func (e *IAMRolePolicyAttachment) Filter() error {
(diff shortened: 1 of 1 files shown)


der-eismann (approved):

kazimierzbudzyk:
@bjoernhaeuser thanks for approving my other PR! Would you mind taking a look at this one as well? It's a simple change but definitely quite annoying to run into it when dealing with us-gov account.

svenwltr (approved):

PR #829, opened 2022-07-08
Fixed version lock on docker command.

ThePrankMonkey:
While attempting to follow the example in your README, I was hitting an issue with the structure of `nuke-config.yml`.

The v2.11.0 version of `aws-nuke` was not marshalling because you now use `account-blocklist` in all of your examples, but that version only accepted the deprecated `account-blacklist`.

If we switch to running the `:latest` image, it'll work with either version of the blocklist key.


der-eismann:
Hey @ThePrankMonkey, thanks for this PR! We won't suggest that people use `latest`, because the main branch can be unstable at times. But I created #891 because the issue is totally valid.

PR #836, opened 2022-07-27
Improve eventbridge resource cleanup

alexzum:
+ Delete targets, rules for any bus. 
+ Delete custom buses.

### DIFF
--- a/resources/cloudwatchevents-buses.go
+++ b/resources/cloudwatchevents-buses.go
@@ -1,8 +1,6 @@
 package resources
 
 import (
-	"fmt"
-
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/cloudwatchevents"
 )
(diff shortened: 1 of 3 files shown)


resources/cloudwatchevents-buses.go
```diff
@@ -0,0 +1,50 @@
...
+
+func (bus *CloudWatchEventsBus) Remove() error {
+	_, err := bus.svc.DeleteEventBus(&cloudwatchevents.DeleteEventBusInput{
+		Name: bus.name,
+	})
+	return err
+}
+
+func (bus *CloudWatchEventsBus) String() string {
+	return fmt.Sprintf("Bus: %s", *bus.name)
```
svenwltr, line 49:
```suggestion
	return *bus.name
```

Please just use the plain bus name, since `String()` is used for filters and having to prepend `Bus: ` might be confusing to users.

alexzum: @svenwltr Thank you for the feedback.  I added a second commit that returns the plain bus name.

PR #840, opened 2022-08-05
[Feat/Addition] Add resource for EC2SpotInstanceRequest

pww217:
First, thanks for the tool, and in advance for taking a look at this PR. 

When using this on our own infra we found that it failed to delete EC2 Spot Instance Requests, which we used for ad-hoc instances and not for fleets.

What this meant is that especially if using persistent spot types, the instances would be terminated by aws-nuke but simply restart themselves because their request was still present.

I'm not 100% sure this is how to solve this problem, so please review. Go is not my strong suite, but my basic thought was that it simply was a resource that wasn't evaluated by the application.

It passed CI and Testing here:

https://github.com/pww217/aws-nuke/runs/7694603979?check_suite_focus=true

I tried to build and test my own image but had some trouble. If that's required I can try to make it work.

### DIFF
--- a/resources/ec2-spot-instance-request.go
+++ b/resources/ec2-spot-instance-request.go
@@ -1,0 +1,66 @@
+package resources
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+)
+
+type EC2SpotInstanceRequest struct {
+	svc   *ec2.EC2
+	id    string
+	state string
+}
+
+func init() {
+	register("EC2SpotInstanceRequest", ListEC2SpotInstanceRequests)
+}
+
+func ListEC2SpotInstanceRequests(sess *session.Session) ([]Resource, error) {
+	svc := ec2.New(sess)
+
+	resp, err := svc.DescribeSpotInstanceRequests(nil)
+	if err != nil {
+		return nil, err
+	}
+
+	resources := make([]Resource, 0)
+	for _, config := range resp.SpotInstanceRequestConfigs {
+		resources = append(resources, &EC2SpotInstanceRequest{
+			svc:   svc,
+			id:    *config.SpotInstanceRequestId,
+			state: *config.SpotInstanceRequestState,
+		})
+	}
+
+	return resources, nil
+}
+
+func (i *EC2SpotInstanceRequest) Filter() error {
+	if i.state == "cancelled" {
+		return fmt.Errorf("already cancelled")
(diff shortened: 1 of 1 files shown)


resources/ec2-spot-instance-request.go
```diff
@@ -0,0 +1,64 @@
...
+func (i *EC2SpotInstanceRequest) Filter() error {
+	if i.state == "cancelled" {
+		return fmt.Errorf("already cancelled")
+	}
+	return nil
+}
+
+func (i *EC2SpotInstanceRequest) Remove() error {
+	params := &ec2.CancelSpotInstanceRequestsInput{
+		// TerminateInstances: aws.Bool(true),
```
der-eismann, line 48: What is this line supposed to do? I can't see the property in the struct [here](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#CancelSpotInstanceRequestsInput)

resources/ec2-spot-instance-request.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (i *EC2SpotInstanceRequest) String() string {
+	return i.id
+}
```
der-eismann, line 64: Please also add properties and tags like [here](https://github.com/rebuy-de/aws-nuke/blob/main/resources/ec2-instances.go#L110:L123) with some [properties the SDK provides](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#SpotInstanceRequest) you might find useful when filtering.

resources/ec2-spot-instance-request.go
```diff
@@ -0,0 +1,64 @@
...
+	register("EC2SpotInstanceRequest", ListEC2SpotInstanceRequests)
+}
+
+func ListEC2SpotInstanceRequests(sess *session.Session) ([]Resource, error) {
+	svc := ec2.New(sess)
+
+	resp, err := svc.DescribeSpotInstanceRequests(nil)
+	if err != nil {
+		return nil, err
+	}
```
der-eismann, line 26: This only returns one page and might not work for people who have many requests. You need to add a paginator like [here](https://github.com/rebuy-de/aws-nuke/blob/main/resources/ec2-instances.go#L45:L51) so that it actually gets all spot instance requests.

der-eismann (dismissed):
Hey @pww217, thanks for your PR! To be honest I'm not 100% sure that this is the correct solution. Maybe it would make sense to filter only for persistent requests? I could imagine that cancelling spot requests created by an autoscaling group before the group itself is deleted would lead the group to create new requests... Please implement the changes I mentioned in the code and then we need to test it.

PR #847, opened 2022-09-01
feat/logo for README

dd84ai:
It looks like really fitting :) found at some tutorial https://dev.to/aws-builders/lets-nuke-aws-account-3j7k


der-eismann:
Hey @dd84ai, thanks for your contribution! Unfortunately we have to reject this, mainly because of copyright reasons - we can't just steal artwork from other sites. They spend their time and creativity on that and we have to respect that :slightly_smiling_face: 
In addition we are not sure if a logo adds any value to the project. And even if the project carries the name, with the current world events happening we don't really want to have a picture of a nuke on our page. But feel free to contribute some code in the future!

PR #878, opened 2022-11-08
Add tag filtering to OpenID connect provider

leighpascoe:
Added ability to filter on the tags of the OpenID connect provider


leighpascoe:
I've tested it locally with the following config
```yaml
---
regions:
- "global" # This is for all global resource types e.g. IAM
- "us-east-1"
- "us-east-2"
- "us-west-1"
- "us-west-2"

account-blocklist:
- "999999999999"


resource-types:

  targets:
  - IAMOpenIDConnectProvider


accounts:
  "123456123123":
    filters:
      IAMOpenIDConnectProvider:
      - property: tag:TempTag
        type: regex
        value: "[\\d]+"
        invert: true
```

der-eismann (approved):

Issue #902, opened 2022-12-29
IAMRole delete failed


optplx:
Hello,

using aws-nuke `latest` docker image (with ID `312d0ebbfe4d`).

I am trying to delete IAM roles using a configuration like below (simplified here for the issue)
```
resource-types:
  targets:
  - IAMRole

presets:
  ignore-aws-nuke:
    filters:
      IAMRole:
      - property: "tag:Ignore-aws-nuke"
         type: exact
         value: true

accounts:
  "xxxxxx":
    presets:
    - ignore-aws-nuke
```

When I run aws-nuke, I get `DeleteConflict: Cannot delete entity, must detach all policies first.`. 

If I understand correctly, I should add filters for all corresponding policy attachments? This is not really easy for my use case, since I will have a lot of filters and each new Role filter should be completed by a policy attachments filter. This feels highly error prone to me

So my question are:
- do I understand correctly?
- if yes would it be acceptable to implement a `force-delete` (or similar) to be able to delete the attachments of the listed IAM roles? I would be glad to discuss a solution with you and work on a PR if so

thanks :)

svenwltr:
Yes I think you do understand it correctly.

> if yes would it be acceptable to implement a force-delete (or similar) to be able to delete the attachments of the listed IAM roles?

The problem here is that aws-nuke is designed to completely ignore dependencies and just retries to delete everything until it succeeds. I am not sure how to solve this without introducing dependencies.

optplx:
I fully understand your concerns, and indeed when pushing the code I was thinking the same about the dependencies. I have proposed a feature flag to make it explicit, but still the dependency will be here.

But imho, not being able to delete IAM roles without having config for attachments makes the tool hardly usable in my context.

If it is not acceptable for you, I think that I will have to live with my own fork :/ (also I will probably have other use-cases with dependencies, that could be worked-around more or less easily. e.g see https://github.com/rebuy-de/aws-nuke/pull/907)

optplx:
I have tested a bit more around the roles and attachments, I do think that those 2 specific resources are tightly coupled, and that mixing those 2 in some way is inevitable. E.g in the current codebase, for the IAMRoleAttachments code, we start by listing the roles. 

This allows for fine filtering of the attachments, but it also has some downsides:
- duplicated config between the roles and attachments, with increased risk of human error
[...]

PR #903, opened 2022-12-30
feat: add force-delete-iam-roles feature flag

optplx:
this is a proposal to address https://github.com/rebuy-de/aws-nuke/issues/902

The solution in a few words:
- this is a new feature flag ` force-delete-iam-roles`
- if true, the removal of a role will first trigger the list and removal of all attached policies (inline and managed), and all instance profiles

Happy to discuss it further, I have tested it on my own environment with success: the first iteration may fail because they happen too soon after the removal of the policies or instance profiles, but with the retries implemented it eventually succeeds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -422,6 +422,7 @@ feature-flags:
     EC2Instance: true
     CloudformationStack: true
   force-delete-lightsail-addons: true
+  force-delete-iam-roles: true
 ```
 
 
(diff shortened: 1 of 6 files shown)


optplx:
hello @svenwltr, any chance this feature will be considered, or should I just close the issue?

PR #904, opened 2022-12-30
feat: add HasTargets property in elbv2 target-groups

optplx:
This adds a property on elbv2 target-groups, to be able to delete unused target groups.
It does add a call per target group, but I did not find any other way of doing so.

What do you think?

example usage 
```
    filters:
      ELBv2TargetGroup:
      - property: HasTargets
        type: exact
        value: true
```


optplx:
actually closing for now, I have to address a few issues before. I will re-open later

PR #907, opened 2022-12-31
feat: add IsEKSManaged property for all elbs and target groups

optplx:
This adds a new property `IsEKSManaged` to the elb, elbv2 and target groups.  

The ALB controller automatically adds a tag `elbv2.k8s.aws/cluster` to those resources  (as per the aws-load-balancer-controller [documentation](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/ingress/annotations/#resource-tags))

Concerning the classic ELBs, they are handled as of today by the in-tree controller and not by the ALB controller. Hence the tag is different, which explains my second commit. I decided to let the other filter as well, in case the ALB controller would take over the management of the classic ELBs at some point.

I am not sure about my implementation though, because it does implicitly bring EKS things into load balancers resources (it would for example require the `eks:ListClusters` permission to someone wanting to manage load balancers only).

However this would be very useful for us: in our context we want to keep all resources deployed via the ALB controller of any EKS cluster alive. When deleting a cluster, it often happens that we have orphan resources.
I'd like to avoid hardcoding the clusters name in the nuke config, because I will 100% sure forget to add or remove a name one day :/

What do you think?

### DIFF
--- a/resources/eks-clusters.go
+++ b/resources/eks-clusters.go
@@ -1,6 +1,7 @@
 package resources
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -51,6 +52,18 @@ func ListEKSClusters(sess *session.Session) ([]Resource, error) {
 		params.NextToken = resp.NextToken
 	}
 	return resources, nil
+}
+
+func mapEKSClusters(sess *session.Session) (map[string]bool, error) {
+	eksClusters, err := ListEKSClusters(sess)
+	if err != nil {
+		return nil, err
+	}
+	eksClustersMap := make(map[string]bool)
+	for _, cl := range eksClusters {
+		eksClustersMap[fmt.Sprintf("%v", cl)] = true
+	}
+	return eksClustersMap, nil
 }
 
 func (f *EKSCluster) Remove() error {
--- a/resources/elb-elb.go
+++ b/resources/elb-elb.go
@@ -1,6 +1,7 @@
 package resources
 
 import (
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws/session"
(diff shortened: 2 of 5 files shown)


svenwltr (dismissed):

svenwltr:
If I see this correctly you can already achieve this with the aws-nuke config, right?

optplx:
you can more or less: we have clusters created and deleted quite often, for experiment purposes. Hence, this would require to update the nuke everytime, which would be painful. This does the same, but dynamically

optplx:
(force-pushed to rebase on main)

svenwltr:
Hello. Sorry for the late answer.

A filter config like this would work and does not require to change the nuke config every time:

```yaml
      ELB:
        - property: tag:elbv2.k8s.aws/cluster
          type: glob
          value: '*'
      ELBv2:
        - property: tag:elbv2.k8s.aws/cluster
          type: glob
          value: '*'
      ELBv2TargetGroup:
        - property: tag:elbv2.k8s.aws/cluster
          type: glob
          value: '*'
```

Or is there something I am missing?

optplx:
actually it would not work, because when we destroy a cluster, the load balancers remain and they are tagged with `elbv2.k8s.aws/cluster` set to the value of the old cluster that has been destroyed.

What would work would be to update the nuke config file to list all the clusters alive, e.g:
```
      ELB:
        - property: tag:elbv2.k8s.aws/cluster
          type: regex
          value: (clusterA|clusterB|clusterC)
      ELBv2:
        - property: tag:elbv2.k8s.aws/cluster
          type: regex
          value: (clusterA|clusterB|clusterC)
      ELBv2TargetGroup:
        - property: tag:elbv2.k8s.aws/cluster
          type: regex
          value: (clusterA|clusterB|clusterC)
```

This would require a lot of work because we have a lot of AWS accounts, with clusters created/destroyed regularly

optplx:
(fixing conflicts)

optplx:
hello @svenwltr any news on this one? :)

Issue #946, opened 2023-02-27
Main Route Tables should be filtered out


andreadecorte:
As per [AWS docs](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Route_Tables.html#main-route-table), we can't delete main route tables, so those should be filtered out as we do for the default resources

der-eismann:
Fixed by #953

Issue #785, opened 2022-04-27
Getting dependency violation with VPC, while deleting from nuke


ukreddy-erwin:
Error I got:

```
us-west-2 - EC2VPC - vpc-043d412c4f7070315 - [ID: "vpc-043d412c4f7070315", IsDefault: "false", tag:Name: "ourproduct-dev-rm-969e-vpc"] - failed
us-west-2 - EC2RouteTable - rtb-0c36016642d3c2a57 - [DefaultVPC: "false"] - failed

Removal requested: 0 waiting, 2 failed, 5 skipped, 0 finished

us-west-2 - EC2VPC - vpc-043d412c4f7070315 - [ID: "vpc-043d412c4f7070315", IsDefault: "false", tag:Name: "ourproduct-dev-rm-969e-vpc"] - failed
us-west-2 - EC2RouteTable - rtb-0c36016642d3c2a57 - [DefaultVPC: "false"] - failed

Removal requested: 0 waiting, 2 failed, 5 skipped, 0 finished

us-west-2 - EC2VPC - vpc-043d412c4f7070315 - [ID: "vpc-043d412c4f7070315", IsDefault: "false", tag:Name: "ourproduct-dev-rm-969e-vpc"] - failed
us-west-2 - EC2RouteTable - rtb-0c36016642d3c2a57 - [DefaultVPC: "false"] - failed

Removal requested: 0 waiting, 2 failed, 5 skipped, 0 finished

ERRO[0057] There are resources in failed state, but none are ready for deletion, anymore.

us-west-2 - EC2VPC - vpc-043d412c4f7070315 - [ID: "vpc-043d412c4f7070315", IsDefault: "false", tag:Name: "ourproduct-dev-rm-969e-vpc"] - failed
ERRO[0057] DependencyViolation: The vpc 'vpc-043d412c4f7070315' has dependencies and cannot be deleted.
        status code: 400, request id: 822ec66c-0a20-43d6-ba22-0861ee8f0970
us-west-2 - EC2RouteTable - rtb-0c36016642d3c2a57 - [DefaultVPC: "false"] - failed
ERRO[0057] DependencyViolation: The routeTable 'rtb-0c36016642d3c2a57' has dependencies and cannot be deleted.
        status code: 400, request id: 17877fae-4b05-4e3b-a08c-9ecc721a1ca0
Error: failed
```

The config file:

```
regions:
- us-west-2
- us-west-1

account-blocklist:
- "999999999999" # production

[...]

PR #953, opened 2023-03-03
Filter main route tables

leighpascoe:
Description

AWS does not allow you to delete main route tables, see [AWS docs](https://docs.aws.amazon.com/vpc/latest/userguide/WorkWithRouteTables.html#DeleteRouteTable). 
Aws-nuke wants to delete a VPC, it expects the main route table to be deleted first, the main route table cannot be deleted, therefore the VPC fails to delete due to a dependency error.

Proposed Solution
After Describing the route tables, check for any `Associations` where `Main` is `true`. There may be zero to many Associations. If at least one `main` Association is found on the Toute Table, then the Route Table will will not be added to the returned collection of Route Tables

I have tested this on VPC with just a main route table, as well as VPC's with multiple route tables.

Related Issues: https://github.com/rebuy-de/aws-nuke/issues/946 https://github.com/rebuy-de/aws-nuke/issues/785

### DIFF
--- a/resources/ec2-route-tables.go
+++ b/resources/ec2-route-tables.go
@@ -4,6 +4,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/rebuy-de/aws-nuke/v2/pkg/types"
+	log "github.com/sirupsen/logrus"
 )
 
 type EC2RouteTable struct {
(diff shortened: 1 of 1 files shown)


der-eismann:
Hey @leighpascoe, sorry for the late reply. We already have a dedicated `Filter()` function for these things, see [this](https://github.com/rebuy-de/aws-nuke/blob/main/resources/ec2-instances.go#L61-L66) for example. Would be great if you could use this instead of your solution.

leighpascoe:
@der-eismann I've made the update, and tested it. Ready for re-review

leighpascoe:
@svenwltr and @der-eismann Could I please get this re-reviewed. We are looking forward to using the released container instead of building our own forked image

leighpascoe:
@svenwltr, Am I missing anything in this PR?

leighpascoe:
@der-eismann  We have been using aws-nuke on this custom branch for the last 4 months to delete VPC, it's well tested. Could we please get this reviewed and merged?

der-eismann (approved):

bjoernhaeuser (approved):

PR #1014, opened 2023-05-31
Fix date format for ASG creationDate

vboulineau:
#769 introduced a new Time format `2006-01-02 15:04:05.000 -0700 MST` which requires 3 digits for decimal seconds, making date like `2023-05-17 10:01:12.14 +0000 UTC` failing to parse.
Fixing the filter to parse any decimal second.

However the filter should never have been added as it's the property that should be correctly serialized to a string (and not relying on Go default `time.Time` to string print) the same way as it's done for `EC2Instance` for instance.

Leaving the fixed date format just in case it's used somewhere else.


vboulineau:
@der-eismann @svenwltr Would it be possible to have a review on this PR? It basically prevent `aws-nuke` to work for ASG with Date filter and the fix is really simple.

bjoernhaeuser (approved):

der-eismann (approved):

Issue #1024, opened 2023-06-12
Add support for AWS Application Migration Service (MGN) resources


davidferguson-cr:
Please add support for nuking AWS Application Migration Service (MGN) resources.

Service info:
https://aws.amazon.com/application-migration-service/

GO SDK reference:
https://docs.aws.amazon.com/sdk-for-go/api/service/mgn/

joelee:
Pull request issued for MGN support: https://github.com/rebuy-de/aws-nuke/pull/1025

der-eismann:
Fixed by #1025

PR #1025, opened 2023-06-12
New support for Application Migration Service (MGN) resources #1024

joelee:
Removal of all AWS Application Migration Service (MGN) Source Servers and Migration Jobs as requested on #1024


der-eismann:
Hey @joelee, thanks for the PR! However I would kindly ask you to replace `Mgn` with `MGN` and add `Properties()` functions for the resources, since these can hold more data than the `String()` ([example](https://github.com/rebuy-de/aws-nuke/blob/main/resources/cloudfront-distributions.go#L65-L73)).

der-eismann (dismissed):

resources/mgn-jobs.go
```diff
@@ -0,0 +1,76 @@
...
+		JobID: f.jobID,
+	})
+
+	return err
+}
+
+func (f *MGNJob) Properties() types.Properties {
+	properties := types.NewProperties()
+	properties.Set("JobID", f.jobID)
+	properties.Set("Arn", f.arn)
```
bjoernhaeuser, line 66:
```suggestion
	properties.Set("ARN", f.arn)
```

We are unfortunately not consistent with this, though I think `ARN` is the correct one.

resources/mgn-source_servers.go
```diff
@@ -0,0 +1,76 @@
...
+		SourceServerID: f.sourceServerID,
+	})
+
+	return err
+}
+
+func (f *MGNSourceServer) Properties() types.Properties {
+	properties := types.NewProperties()
+	properties.Set("SourceServerID", f.sourceServerID)
+	properties.Set("Arn", f.arn)
```
bjoernhaeuser, line 66:
```suggestion
	properties.Set("ARN", f.arn)
```

Similar thing here.

resources/mgn-jobs.go
joelee: Correction made. Thank you for the review.

bjoernhaeuser (approved):

der-eismann (approved):

PR #1053, opened 2023-07-25
Stopped Executions before deleting step functions state machines

suleman-sohail:
## Description
When we have created state machines in step functions, and have enabled executions within those state machines, we sometimes have trouble in deleting those state machines. This PR fixes that.

### Repro
Create a state machine, with `wait` module (in flow tab, ig), this will have the executions in running state for that duration. Trying to delete state machine would fail because of on-going executions. We have to stop executions, for that state machine.


bjoernhaeuser (dismissed):

resources/sfn-statemachines.go
```diff
@@ -47,6 +47,28 @@ func ListSFNStateMachines(sess *session.Session) ([]Resource, error) {
 }
 
 func (f *SFNStateMachine) Remove() error {
+	params := &sfn.ListExecutionsInput{
+		StateMachineArn: f.ARN,
+	}
+	
+	for{
```
der-eismann, line 54: Can you please run `go fmt` to fix the formatting here?

der-eismann (approved):

svenwltr (approved):

PR #1055, opened 2023-07-26
appconfig: add support for five resources

tbarrella:
Adds support for

- Application
- ConfigurationProfile
- DeploymentStrategy
- Environment
- HostedConfigurationVersion

As part of #1058


bjoernhaeuser (commented):
I think we also need paging for these resources as well

tbarrella:
Mm, anything in particular I need to add/change? I used the `List*Pages` APIs

bjoernhaeuser (approved):
Sorry, i was wrong

der-eismann (approved):

PR #1096, opened 2023-08-30
Improve listing of cloudcontrol resources

der-eismann:
- Filtering by AWS prefix
- Do it twice for FullyMutable & Immutable (just means can't be updated, but can be deleted


svenwltr (approved):
Be aware that querying the cloud control data is very slow.

bjoernhaeuser (approved):

PR #1193, opened 2024-02-20
Add ELB Listener Rules object

vboulineau:
Adding ELB Listener Rule resource.
It cannot be cleaned up just by using Cloud Control as it requires to list ELB -> Listener -> ListenerRule

### DIFF
--- a/resources/elbv2-listenerrule.go
+++ b/resources/elbv2-listenerrule.go
@@ -1,0 +1,141 @@
+package resources
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/elbv2"
+	"github.com/rebuy-de/aws-nuke/v2/pkg/types"
+	"github.com/sirupsen/logrus"
+)
+
+var elbv2ListenerRulePageSize int64 = 400 // AWS has a limit of 100 rules per listener
+
+type ELBv2ListenerRule struct {
+	svc         *elbv2.ELBV2
+	ruleArn     *string
+	lbName      *string
+	listenerArn *string
+	tags        []*elbv2.Tag
+}
+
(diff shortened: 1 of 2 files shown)


resources/elbv2-listenerrule.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	return nil
+}
+
+func (e *ELBv2ListenerRule) Properties() types.Properties {
+	properties := types.NewProperties().
+		Set("ARN", e.rule.RuleArn)
+	properties.Set("ListenerARN", e.listener.ListenerArn)
+	properties.Set("LoadBalancerName", e.elb.LoadBalancerName)
```
der-eismann, line 131: These can be chained like
```suggestion
		Set("ARN", e.rule.RuleArn).
		Set("ListenerARN", e.listener.ListenerArn).
		Set("LoadBalancerName", e.elb.LoadBalancerName)
```

resources/elbv2-listenerrule.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+var elbv2ListenerRulePageSize int64 = 400 // AWS has a limit of 100 rules per listener
+
+type ELBv2ListenerRule struct {
+	svc      *elbv2.ELBV2
+	elb      *elbv2.LoadBalancer
+	listener *elbv2.Listener
+	rule     *elbv2.Rule
```
der-eismann, line 18: Do we really need to store whole objects here? Looks like we're only using ARNs and names from the resources.

vboulineau: It's indeed not required, I've changed to only store what's necessary.

der-eismann (approved):

bjoernhaeuser (approved):

PR #1224, opened 2024-05-17
allow nuking accounts with prod substring, if you wish

jasonwbarnett:
This program seems to have a lot of safeguards, and I felt hard coded prod substring was too restrictive/prescriptive. I updated the code to allow folks to define their own substring's they wish to block. If none, then don't add any. If `"prod"`, then add it.


jasonwbarnett:
The README.md wasn't updated because I wasn't sure if the maintainers were even receptive to this kind of change. If ya'll are I will work to update the README.md so that it includes these changes.

PR #1248, opened 2024-07-25
fix: skip kms keys if describe is disallowed

mKeRix:
We noticed that aws-nuke will abort listing KMS keys if it does not have permissions to read one of them, which means that a single key with a key policy that does not allow reading is all it takes for all keys to not be nuked anymore. This change updates the behavior so that these keys will be skipped, allowing other keys to still be nuked. This is done because it can be intentional that some keys will not be readable by their key policy, and aws-nuke should be able to handle that.

### DIFF
--- a/resources/kms-keys.go
+++ b/resources/kms-keys.go
@@ -1,11 +1,15 @@
 package resources
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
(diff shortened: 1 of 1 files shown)


ekristen:
@mKeRix this has been implemented via https://github.com/ekristen/aws-nuke/pull/260 - this is now the active fork of aws-nuke.

This project has now been deprecated in favor of [this fork](https://github.com/ekristen/aws-nuke). Sven kindly granted me access to directly answer and close pull requests and issues so that we can notify users if their issues have been addressed or not. Please see the [welcome issue](https://github.com/ekristen/aws-nuke/issues/276) for more information.

PR #1260, opened 2024-09-05
add deprecation notice

svenwltr:
This adds a notice to the README and just states, which unfortunately already was a fact.


bjoernhaeuser (approved):

ekristen:
@svenwltr @bjoernhaeuser thank you for all your hard work on this project. Thanks for letting me take up the reigns. You are welcome to be contributors and even maintainers should you want on my fork. None of this would have been possible without all your hard work and vision that came before.

